PR #11, opened 2015-02-03
Use stream_is_local instead of limit to file://

Slamdunk:
Checking for `file://` limits a lot the possibilities, and it is not totally safe: built-in function is far more reliable.

The main reason for this commit is to allow the use of virtual local stream wrapper in unit-tested project, like [vfsStream](https://github.com/mikey179/vfsStream), in order to not touch real filesystem while testing the creation of PDF with this third-party library.


nicolaasuni:
Manually merged. Thanks.

PR #15, opened 2015-04-08
Added check for if it's same page. To prevent starting on ...

Nicolai-:
...the same page again. Checked this against bug report 948 and 951 where changes to the table methods were done. Also checked all current examples.

This fixed an issue I had were sometime the Table of Content would start on the same page again. It simply skipped the page break. And wrong "second" page on top of the "first" page.


nicolaasuni:
Manually merged with a change. Thanks.

PR #17, opened 2015-04-22
Avoid warning when iterating a non-array variable

jcrombez:
In some cases, $sarr is not an array, and the foreach trigger a warning.
This 3 simple lines prevent this from happening.


nicolaasuni:
Manually merged. Thanks.

PR #18, opened 2015-04-29
Removed qrcode.orig.php

daeMOn63:
Seems like a merge file has been commited. 
Causing a warn from composer update saying : 
```
Warning: Ambiguous class resolution, "QRcode" was found in both "./vendor/composer/tecnick.com/tcpdf/include/barcodes/qrcode.orig.php" and "./vendor/composer/tecnick.com/tcpdf/include/barcodes/qrcode.php", the first will be used.
```


kAworu:
+1

Slamdunk:
:+1:

PR #20, opened 2015-05-16
Avoid error with getimagesize when file URI contains a space

sorich87:
When a remote file has a space in the URL, there's an error because getimagesize requires the space to be encoded to `%20`.

The error message is `failed to open stream: HTTP request failed! HTTP/1.1 505 HTTP Version not supported`.

This pull request fixes it.


nicolaasuni:
manually merged

PR #21, opened 2015-05-19
Unneeded set of y position to old value, added the possibility to again create anchor links within the pdf

Nicolai-:
with <a href="destination_name">Link text</a>... This has been broken for quite a while. Recently upgraded from 6.0.037 to 6.2.6. Just noticed that the links werent working anymore in the pdf.


Nicolai-:
Updated the pull request.
Had an issue with blank pages being inserted, on a manual check. I check for where the "y" position on a page is, so i'm sure there is enough space for content and not just the title for the content. To avoid having title as the last thing on a page.
The issue is that the $this->y in tcpdf.php is reset by a automatic pagebreak, but then afterwards it would be reset to the position which caused the page break,
when i then check the "y" position is last at the page and i therefor also adds an page.

nicolaasuni:
manually merged

PR #23, opened 2015-07-07
Make dependency on mcrypt optional

remicollet:
Hi,

About libmcrypt and php-mcrypt, read http://blog.remirepo.net/post/2015/07/07/About-libmcrypt-and-php-mcrypt

This quick implementation seems to work, minimal test, using example_016.php changed to:

      $pdf->SetProtection(array('print', 'copy'), 'secret', null, 3);

evince and okular correctly ask for password and decryp the generated PDF.

Proably, more test neeeded (especially as I don't understand the need of the substr($x,0,-16))


remicollet:
Thanks!

PR #27, opened 2015-08-24
Add support for compact PDF417 (described http://www.neodynamic.com/Products/Help/BarcodeCF2.0/barcodes/CompactPdf417.htm)

jeffrego:
In our use of this library we have a need to generate PDF417 barcodes with a truncated stop code, trading off extra width for reduced error correction capability (described here: http://www.neodynamic.com/Products/Help/BarcodeCF2.0/barcodes/CompactPdf417.htm).  I have extended the PDF417 class to support this as a new type.


nicolaasuni:
Please note that I am refactoring TCPDF and the barcode library has been moved at:

PR #30, opened 2015-09-14
Use standardized license identifier

valioDOTch:
SPDX has standardized the license identifiers to avoid confusions and allow tool-assisted license checks
https://spdx.org/licenses/


nicolaasuni:
Please note that I am currently refactoring TCPDF and I am not looking into changing the current version until the new one is ready.

valioDOTch:
I now proposed the file change for the respective libraries

PR #32, opened 2015-12-21
Fix composer.json validation

j0k3r:
- License "LGPLv3" is not a valid SPDX license identifier, see http://www.spdx.org/licenses/ if you use an open license. If the software is closed-source, you may use "proprietary" as license.
- The version field is present, it is recommended to leave it out if the package is published on Packagist.


j0k3r:
Well, even if the projet is dead and since the new one isn't still released as stable, why don't you merge these little update?

PR #33, opened 2015-12-22
Fix buggy Byte compaction

claudepache:
Fixes https://sourceforge.net/p/tcpdf/bugs/1034/


nicolaasuni:
The current version of TCPDF will be soon replaced by a new version.
The barcode library has been already ported here:

PR #34, opened 2016-01-12
Normalize Unicode strings before converting them to an array

sebastien-fauvel:
- Fixes PDF rendering issues with combining diacritical marks
- Depends on PECL intl >= 1.0.0 (optional)

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1996,6 +1996,13 @@ class TCPDF_FONTS {
 	 */
 	public static function UTF8StringToArray($str, $isunicode=true, &$currentfont) {
 		if ($isunicode) {
+			static $hasNormalizerSupport = null;
+			if (null === $hasNormalizerSupport) {
+				$hasNormalizerSupport = function_exists('normalizer_normalize') && ("\xC3\x85" === normalizer_normalize("A\xCC\x8A"));
+			}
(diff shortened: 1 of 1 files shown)


sebastien-fauvel:
@nicolaasuni please review & merge. Feel free to move the code to a static method in whatever class fits you best!

nicolaasuni:
I am currently working on a complete refactoring of TCPDF.
The Unicode processing has been moved to a separate library:
https://github.com/tecnickcom/tc-lib-unicode

Please check if your patch is still relevant there.

sebastien-fauvel:
Hi Nicola, could you please provide me a link to the refactored version of TCPDF, if you've already made it public? I don't know if my fix will still be needed, it depends on the usages.

nicolaasuni:
The TCPDF itself is not yet public but the unicode and font libraries are ready:
https://github.com/tecnickcom/tc-lib-unicode-data
https://github.com/tecnickcom/tc-lib-unicode
https://github.com/tecnickcom/tc-lib-pdf-font

sebastien-fauvel:
The fix should be ported to [`Com\Tecnick\Unicode\Convert::strToChrArr`](https://github.com/tecnickcom/tc-lib-unicode/blob/develop/src/Convert.php#L67) then. I will open a PR.

PR #35, opened 2016-01-12
fix loop condition when parsing XREF

lkalinka:
Fix for loop condition reference to this : https://github.com/tecnickcom/tc-lib-pdf-parser/commit/6e35159fb22dc0ac051d5d99ee5ff6fe9db7195a


nicolaasuni:
This project will be soon deprecated in favour of https://github.com/tecnickcom/tc-lib-pdf-parser
Please use the new project.

PR #36, opened 2016-01-27
Update tcpdf_import.php

TheDevilOnLine:
Please don't leave debug code in the stable branch!


nicolaasuni:
This is actually wanted as it is a development file.
In any case, I am currently refactoring TCPDF and I am not looking into changing the current version until the new one is ready.

PR #38, opened 2016-02-09
Improved condition to read external config files and added lib/tcpdf/…

nicraMarcin:
…tcpdf_config.php as subdirectory in project, not globally, to read config file.


nicolaasuni:
I am currently refactoring TCPDF and I am not looking into changing the current version until the new one is ready.

PR #41, opened 2016-05-31
Add custom properties to a document's XMP metadata

jaykravetz:
Hello,

I was looking for a way to add custom properties to a document but couldn't find anything that would work for my use case.

My use case was to add to add custom properties that could be read back by other PDF libraries (In this case, Python's PyPDF2). A plus would be to get Adobe Reader to display those properties as well.

I added some code that's so far working out quite well so far and thought i'd share it.

Any feedback would be well received.

Thank you,
jay

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -1779,6 +1779,13 @@ class TCPDF {
 	protected $custom_xmp = '';
 
 	/**
+	 * Custom document properties inside the XMP data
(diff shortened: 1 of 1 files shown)


nicolaasuni:
Hi, this version has been frozen and the development has been moved to a new repo:
https://github.com/tecnickcom/tc-lib-pdf

PR #43, opened 2016-06-13
BUGFIX: Removing invalid return path of fileGetContents

fcool:
If the server tcpdf runs on has the environment variable REQUEST_URI set, the function fileGetContents replied with an URL and not the file content as expected.
This fix removes this invalid return.

### DIFF
--- a/include/tcpdf_static.php
+++ b/include/tcpdf_static.php
@@ -1841,8 +1841,8 @@ class TCPDF_STATIC {
 	/**
 	 * Wrapper to use fopen only with local files
 	 * @param filename (string) Name of the file to open
-	 * @param $mode (string) 
-	 * @return Returns a file pointer resource on success, or FALSE on error.  
+	 * @param $mode (string)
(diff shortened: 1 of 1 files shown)


filiplikavcan:
Please merge this PR. Thank you.

bruce-thomas:
I also had the same issue, when using the `ImageEps()` method, generating the following **TCPDF error** message :

> No BoundingBox found in EPS/AI file

This is because the method `TCPDF_STATIC::fileGetContents()` is returning the file URL instead of the file content (because my server has the environment variable `SCRIPT_URI` set), thus the file content could not be parsed correctly by the `ImageEps()` method.
When calling `ImageEps()` with the file URL, instead of the file local path, the file content is correctly retrieved, and the error is not raised.

Please, could you consider to merge this PR ?
Thank you

mestremuten:
This seems a duplicate of https://github.com/tecnickcom/TCPDF/pull/62

PR #44, opened 2016-08-26
Add Japanese post card format

MysticDoll:


nicolaasuni:
All Japanese formats are already in tc-lib-pdf.

PR #46, opened 2016-10-04
Barcode - Fixes for PHP7

StefanHefner:
http://php.net/manual/en/migration70.incompatible.php

Negative Bitshifts are throwing a ArithmeticError now.

2 Choices: Prevent negative Bitshifts or catch the Exception  (like here).


SailorMax:
try{}catch(e){} - is too slow.
May be better add "if ($ecl >0)" before shifts?

nicolaasuni:
This project is not supported anymore. All the efforts will be concentrated on the new library:
https://github.com/tecnickcom/tc-lib-pdf
If you just need a barcode library, please have a look a the new version here:

PR #50, opened 2017-01-10
Update tcpdf.php

dardoguidobono:
fix problem on trying to set array value in a string

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -19749,7 +19749,7 @@ Putting 1 is equivalent to putting 0 and calling Ln() just after. Default value:
 				break;
 			}
 			case 'a': {
-				$this->HREF = '';
+				$this->HREF = array();
 				break;
 			}
 			case 'sup': {


alecsammon:
Can we get this merged in - without this fix then the library fails on PHP 7.1

Aware that the project is no longer maintained - but this should be an easy patch to apply.

Thanks!

jblotus:
yes would be nice to merge this in - for now I'm forking to deal with this issue

amakhrov:
@nicolaasuni If you could merge and publish this fix, it would be just wonderful.

Happy Valentine's! :)

kristian-b:
A fix for that would be very nice for me too.

hoopyfrood:
+1

yakobe:
Hi @nicolaasuni

yakobe:
Hi @nicolaasuni . I believe you are the owner of this repo. Would it be possible to merge this in so we can continue to use this lib?

I understand that there is a [new version in development](https://github.com/tecnickcom/tc-lib-pdf) but there does not seem to be any documentation about how to use it or when it will be stable. Or am i looking in the wrong place?

yakobe:
Thanks 👍

kristian-b:
Thank you @nicolaasuni !

PR #51, opened 2017-01-14
Add Korean Character Support in Output Filename

Mabin-J:
Output Filename Character Restriction is too tight.


divinity76:
i have a better idea, see  https://github.com/tecnickcom/TCPDF/pull/59

nicolaasuni:
The output file name are ASCII only for wide compatibility.
New features will be considered in tc-lib-pdf

PR #55, opened 2017-04-10
Added pdf/x-3 support

4NdIb:
I have added pdf/x-3 support as described on this page: http://www.pdfx3.org

It also works with FPDI extension when you include valid pdf/x-3 files.

Hope you enjoy.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -1757,6 +1757,12 @@ class TCPDF {
 	 */
 	protected $pdfa_mode = false;
 
+    /**
+     * If true set the document to PDF/X mode.
+     * @protected
+     */
+    protected $pdfx_mode = false;
+
 	/**
 	 * Document creation date-time
 	 * @protected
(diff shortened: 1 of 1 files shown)


nicolaasuni:
I will add this to the new library as the current PR breaks backward compatibility and this version is frozen.

jhonnymakis:
tcpdf.php added the code below to fix a issue with custom fonts; After line 4258

The name the font was **nexaslab** and the script was sending with ' like this  **'nexaslab'.php** instead  **nexaslab.php** that's why it can't find the font file. 

`$family = str_replace("'", '', $family);`

tcpdf.php
```diff
@@ -1831,11 +1837,11 @@ class TCPDF {
 	 * @param $unicode (boolean) TRUE means that the input text is unicode (default = true)
 	 * @param $encoding (string) Charset encoding (used only when converting back html entities); default is UTF-8.
 	 * @param $diskcache (boolean) DEPRECATED FEATURE
-	 * @param $pdfa (boolean) If TRUE set the document to PDF/A mode.
+	 * @param $pdfa (string) If pdfa set the document to PDF/A mode, if pdfx set the document to PDF/X mode or none.
```
kyborek: The parameter name should now be $pdfmode
It would be helpful if specifically documented that the mode is PDF/X-3 as others could assume it is PDF/X-1 for example

nicolaasuni:
Thank for your contribution.

PR #60, opened 2017-06-14
fix Undesired mouseover effect on links in PDF on Chrome Pdf Viewer

spipu:
It is to fix a pb on Chrome Pdf Viewer.

Lots of pdf lib have the pb : https://github.com/mpdf/mpdf/issues/283

I know that now this package is deprecated, but since the new lib is not ready, it could be cool to manage bug fix on this version ;)


spipu:
i close this one because i use a bad git user, sorry :(

PR #62, opened 2017-06-14
Add url as an alternative path instead of returning directly

tacovandenbroek:
When `TCPDF_STATIC::fileGetContents` is called with a relative path while `$_SERVER['SCRIPT_URI']` is set, this method _always_ used the web-url to retrieve the contents. This fix will just add the url as one of the alternatives to be checked later in the method.


mestremuten:
@nicolaasuni I know this library is no longer being developed, in the sense that no new features will be added, but while the new version is not near stable, imho bugs like this could be fixed.

This bug prevents using ImageSVG and ImageEps methods with file path as argument.

PR #63, opened 2017-06-14
Support ctg files that have not been lowercased

tacovandenbroek:
In our (and probably) other legacy codebase previously created fonts using ufm have not all been lowercased. This pull request will support both the lowercased files and the case sensitive files.

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1541,16 +1541,28 @@ class TCPDF_FONTS {
 	 * @public static
 	 */
 	public static function getFontFullPath($file, $fontdir=false) {
-		$fontfile = '';
-		// search files on various directories
-		if (($fontdir !== false) AND @file_exists($fontdir.$file)) {
-			$fontfile = $fontdir.$file;
-		} elseif (@file_exists(self::_getfontpath().$file)) {
-			$fontfile = self::_getfontpath().$file;
-		} elseif (@file_exists($file)) {
-			$fontfile = $file;
+		// Look for both the original file name and the lowercased variant
+		$alternatives = [
+			$file,
+			strtolower($file)
+		];
+
+		foreach ($alternatives as $alternative) {
+			// search files on various directories
+			if (($fontdir !== false) AND @file_exists($fontdir.$alternative)) {
+				return $fontdir.$alternative;
+			}
+
+			if (@file_exists(self::_getfontpath().$alternative)) {
+				return self::_getfontpath() . $alternative;
+			}
+
+			if (@file_exists($alternative)) {
+				return $alternative;
+			}
 		}
-		return $fontfile;
+
+		return '';
 	}
 
 
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -8994,7 +8994,7 @@ class TCPDF {
 			// Embed CIDToGIDMap
 			// A specification of the mapping from CIDs to glyph indices
 			// search and get CTG font file to embedd
-			$ctgfile = strtolower($font['ctg']);
+			$ctgfile = $font['ctg'];
 			// search and get ctg font file to embedd
 			$fontfile = TCPDF_FONTS::getFontFullPath($ctgfile, $fontdir);
 			if (TCPDF_STATIC::empty_string($fontfile)) {


divinity76:
rather than supporting "all lowercase" and "exact" filename, i suggest making a completely filename-case-insensitive search, perhaps like this (UNTESTED):

```php
public static function getFontFullPath($file, $fontdir = false) {
	// filename-case-insensitive search for font file in current dir and all fonddirs
	if (empty ( $fonddir )) {
		$fontdir = [ 
				'.' 
		];
	} else {
		$fonddir [] = '.';
	}
	
	$file = basename ( strtolower ( trim ( $file ) ) );
	foreach ( $fontdir as $fd ) {
		$di = new DirectoryIterator ( $fd );
		foreach ( $di as $file ) {
			if (strtolower ( trim ( $file->getBasename () ) ) === $file) {
				return $file->getRealPath ();
			}
		}
	}
	return '';
}
```

divinity76:
rather than supporting "all lowercase" and "exact" filename, i suggest making a completely filename-case-insensitive search, perhaps like this (UNTESTED):

```php
public static function getFontFullPath($file, $fontdir = false) {
	// filename-case-insensitive search for font file in current dir and, and optionally fontdir
	if (empty ( $fontdir )) {
		$fontdir = [ 
				'.' 
		];
	} else {
		$fontdir = [ 
				$fontdir,
				'.' 
		];
	}
	
	$file = basename ( strtolower ( trim ( $file ) ) );
	foreach ( $fontdir as $fd ) {
		$di = new DirectoryIterator ( $fd );
		foreach ( $di as $file ) {
			if (strtolower ( trim ( $file->getBasename () ) ) === $file) {
				return $file->getRealPath ();
			}
		}
	}
	return '';
}
```

tacovandenbroek:
I see why you would think that a case insensitive search might be interesting. However, I don't see any reason why a filename would be formatted in any other way than the original filename or completely lowercased. Besides that, your solution iterates over several directories (the path self::_getfontpath() is missing), potentially containing many font files, this could be a unwanted performance hit.

divinity76:
`I don't see any reason why a filename would be formatted in any other way than the original filename or completely lowercased`

well, i do. what makes you think the only CaSe bugs people have, is having font file names all lowercase? as far as i can see, the only reason people would have a problem with cases on the font files (like on your legacy codebase), is because they migrated from a system where fopen is case-insensitve (like MS Windows), to a system where fopen is case-sensitive (like Linux), and had some previously unnoticed cAsE bugs in their code. i guess the second best solution (short of the best solution, fixing your code) is 
 to mimic the case-insensitive system, which my code does. 

`(the path self::_getfontpath() is missing)`
yup, that's a bug. guess it needs `if(!empty(self::_getfontpath())){$fontdir[]=self::_getfontpath();}` 

`your solution iterates over several directories (the path self::_getfontpath() is missing), potentially containing many font files, this could be a unwanted performance hit.`
agreed. it could be optimized to not do any iterations provided the case is already correct, and fallback to iterate the dirs  if necessary (the case is incorrect). furthermore, it could even be optimized to check if the case bug is that the file is all lowercase first (this is the only case bug your code covers)

divinity76:
... like this (still untested)

```php
public static function getFontFullPath($file, $fontdir = false) {
	// filename-case-insensitive search for font file in current dir and, and optionally fontdir
	$fontdirs = [ ];
	if (! empty ( $fontdir )) {
		$fontdirs [] = $fontdir;
	}
	if (! empty ( self::_getfontpath () )) {
		$fontdirs [] = self::_getfontpath ();
	}
	$fontdirs [] = '.';
	// the fastest path, assuming font cases are 100% correct:
	foreach ( $fontdirs as $path ) {
		if (file_exists ( $path . $file )) {
			return $path . $file;
		}
	}
	// 2nd fastest path,
	// special optimization for the bug "the font filename is all lowercase:
	$file = strtolower ( $file );
	foreach ( $fontdirs as $path ) {
		if (file_exists ( $path . $file )) {
			return $path . $file;
		}
	}
	// slowest path: completely case-insensitive filename search:
	foreach ( $fontdirs as $fd ) {
		$di = new DirectoryIterator ( $fd );
		foreach ( $di as $file ) {
			if (strtolower ( $file->getBasename () ) === $file) {
				return $file->getRealPath ();
			}
		}
	}
	// failed to find the font.
	return '';
}
```

nicolaasuni:
Thank for your contribution.

PR #68, opened 2017-08-02
"Non well formed numeric value" error occurs in PHP 7

csears123:
The $size variable in the SetFontSize function is supposed to be a float type, but a string can be passed in like "14px" causing an error in PHP 7.  Re-casting the variable is a simple solution to fix the variable type if it was incorrectly passed in.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -4453,6 +4453,7 @@ class TCPDF {
 	 * @see SetFont()
 	 */
 	public function SetFontSize($size, $out=true) {
+		$size = (float)$size;
 		// font size in points
 		$this->FontSizePt = $size;
 		// font size in user units


sayoki:
Hi,

I should have signed it.


Greetings,

Sjors

2017-08-02 17:51 GMT+02:00 CLAassistant <notifications@github.com>:

divinity76:
the warning is correct, and arguably, "14px" should throw an InvalidArgumentException, because it cannot be losslessly converted to float. 

for PHP5 support, could do something like 
```php
$size = filter_var($size,FILTER_VALIDATE_FLOAT);
if(false===$size){
    throw new \InvalidArgumentException('Argument 1 must be of the type float');
}
```
and when PHP5 support is ditched, the equivalent PHP7 solution is to simply change the input paramaters to 
```php
  	public function SetFontSize(float $size, bool $out=true) {
```

sayoki:
Can we get a merge?

csears123:
In response to @divinity76 simply adding the type declaration "float" for the $size variable in the input parameter will not resolve the error "non well formed numeric value encountered" in PHP 7 because the functions that execute SetFontSize are still passing in a string "14px" (at times) and PHP will throw the same error causing the application to fail.  Additionally, when adding type declarations to the input parameters (starting with PHP 5) if a value other than 'null' is assigned a PHP fatal error will occur "Default value for parameters with a class type can only be NULL".  So the suggestion of "bool $out=true" for the second input parameter will also fail.  
I would recommend leaving the initial solution in place, because the desired result is to gracefully handle the variables passed in **without** throwing an error.  The more robust solution would be to address the issue upstream with the functions that call SetFontSize, and sanitize the inputs before passing them to this function, however since this repository is being considered "obsolete" and "will not receive additional development or support", the initial solution I proposed should be sufficient.

divinity76:
quote: `because the functions that execute SetFontSize are still passing in a string "14px" ` - that is incorrect. at least in php7 (not sure how this worked back in php5), if the input has a type, the variable is guaranteed to be that type inside the function (the exception to this is if the default value is explicitly NULL, then the variable is guaranteed to be the declared type, or be NULL.)
```php
<?php
declare(strict_types=0);
f('14px');
function f(float $v){
    // in here, $v is __guaranteed__ to be a float.
    var_dump($v);
}
```

[in this code, the f function passes a true float to var_dump, float(14)](https://3v4l.org/rK6mt)

quote: `So the suggestion of "bool $out=true" for the second input parameter will also fail.` - that's also incorrect. it's true, this is what happens in PHP5, as PHP5 think you mean the `class bool`, not an actual boolean, but PHP7 added support for bool input types (along with `int`, `float`, and `string`), so as i said, `when PHP5 support is ditched`, that's a valid (and pretty) approach, but not until then.


quote: `gracefully handle the variables passed in without throwing an error`, i never liked this approach anyway.  when the programmer is doing something wrong, throw an error! let him know that he's doing something wrong, and should fix his code.

csears123:
The default mode for type declarations is coercive, which will produce the PHP Notice of "A non well formed numeric value encountered" in PHP 7.1.7.  Within my application, there is custom error handling using 'set_error_handler()' that will trow an exception for these types of errors, preventing the application from continuing within the try-catch block when using TCPDF.  Enabling strict mode (declare statement below), will then throw a TypeError exception, because it does not match the exact type specified.  The application will still fail at this point.
`declare(strict_types=1);`

For the second parameter, you are correct, I used 'boolean' instead of 'bool' for the parameter type, and using 'bool' is accepting the default value of true.  However it would be ideal to have a solution that applies to both PHP 5 and PHP 7, to not alienate older applications.  

The initial proposed solution will support both PHP 5 and PHP 7 and not have any dependencies on which version is used.

sayoki:
This TCPDF build keeps generating errors.
The error I see at the moment is: 

Undefined index: [file] => tcpdf/tcpdf.php [line] => 19458 )

PR #69, opened 2017-08-10
Fix tcpdf_addfont.php Bug

zeldajay:
PHP in different operating systems installed location is not the same, so can not write so.


zeldajay:
merge pull requests.

PR #71, opened 2017-09-12
Update tcpdf_static.php

madaxel:
add URL to alternative array instead of returning directly


madaxel:
already exists https://github.com/tecnickcom/TCPDF/pull/62

PR #72, opened 2017-09-12
MCRYPT is deprecated in PHP 7.1

robernet:


nicolaasuni:
This has been already fixed in the new tc-lib-pdf library (incomplete).

it-can:
in php 7.2 mcrypt is removed and will thrown an error...

PR #74, opened 2017-09-21
Refactor deprecated each() calls to foreach()

emielmolenaar:
See http://php.net/each. This function has been DEPRECATED as of PHP 7.2.0. Relying on this function is highly discouraged.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -12582,7 +12582,7 @@ class TCPDF {
 		$k = $this->k;
 		$this->javascript .= sprintf("f".$name."=this.addField('%s','%s',%u,[%F,%F,%F,%F]);", $name, $type, $this->PageNo()-1, $x*$k, ($this->h-$y)*$k+1, ($x+$w)*$k, ($this->h-$y-$h)*$k+1)."\n";
 		$this->javascript .= 'f'.$name.'.textSize='.$this->FontSizePt.";\n";
-		while (list($key, $val) = each($prop)) {
+		foreach($prop as $key => $val) {
 			if (strcmp(substr($key, -5), 'Color') == 0) {
 				$val = TCPDF_COLORS::_JScolor($val);
 			} else {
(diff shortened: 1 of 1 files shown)


it-can:
@nicolaasuni can you merge this? TCPDF is not working with PHP 7.2 at the moment...

acirinelli:
Can we get someone to merge please? 👍

fezfez:
Need the same, my app cannot move to php 7.2 until it's merged !

hotdesc:
Merge plz?

fezfez:
Is @nicolaasuni alive ?

nicolaasuni:
When I will have some time I will update the new version as this one is frozen.

semaex:
Same problem, my app cannot move to php 7.2

gargoyle:
If you are using composer to install on a system with a compatible `patch` command (default one on ubuntu works for me), then you can apply the following "sticky plaster" fix as a temporary measure...

1.) Download this pull requests diff by adding `.patch` to the end of the url. GitHub will re-write it to this:- https://patch-diff.githubusercontent.com/raw/tecnickcom/TCPDF/pull/74.patch

2.) Save the file along side your `composer.json` file.

3.) Edit your `composer.json` and add the following snippet:-
```
"scripts": {
        "post-install-cmd": "patch vendor/tecnickcom/tcpdf/tcpdf.php tcpdf.patch"
    }
```

If you already have a `scripts` section, it should be obvious how you can include just this post install command.

4.) Run `composer install`.

(Note: `composer update` will not trigger the patch. I'm not sure why composer doesn't supply a hook to cover ether an update or an install)

Sedaj:
Try this solution to apply @emielmolenaar PR right away: 

 1. Go to your composer.json and change in require key:
    `"tecnickcom/tcpdf": "dev-master"`
2. Update repositories key like:
    `"repositories": [`
     `      {`
     `       "type": "vcs",`
    `        "url": "https://github.com/emielmolenaar/TCPDF"`
    `    } `
    ` ]`
3. Run composer update

emielmolenaar:
@nicolaasuni which new version?

fezfez:
I don't understand "I will update the new version as this one is frozen.".

This MR is about a bug fix, this is not a major version, neither a minor but a patch, it can be apply to **any** major.minor version that's have this bug without breaking the semver rule.
There is no dev branches (so no new version incoming), so i don't understand why it's a big deal to merge this.

You will be thanks-fulled by dozen of people if you will merge this.

Thanks in advance.

iquito:
I don't quite understand why this is not merged - it is a super simple bugfix which does not change anything about the library, it just fixes it for PHP 7.2. Creating a completely new library instead does not change that people will experience this issue when moving to PHP 7.2 - I did too, which is why I am here.

In addition, the new library does not even exist yet in a stable form. This basically means TCPDF is completely abandoned right now for no good reason.

gargoyle:
Yeah. It's normally best practice to deprecate the old version AFTER the new one has a stable release. 

There seems to be plenty of people here who seem to be sensible enough and hopefully willing to help - perhaps you could consider giving someone merge capability to maintain this version from a serious bug fix perspective while the finishing touches are being put to the new version.

nicolaasuni:
Unfortunately I wasn't able to find the time to complete the new tc-lib-pdf, so I will review and merge a bunch of PRs in the old TCPDF soon.

emielmolenaar:
Thank you @nicolaasuni :tada: !

PR #75, opened 2017-09-28
* Add 3 Droid fonts which were there in previous versions

plamendp:
These are fonts we need and found out they were there before, so can we have them back in?


nicolaasuni:
Add binary files poses a licensing issue.
The new tc-lib-pdf library will have a better mechanism to handle fonts.
In the meanwhile you can add fonts case by case.

PR #80, opened 2017-12-12
Handle side padding correctly on writeHTML

NikkiDelRosso:
Previously, the right side padding on HTML cells was set to whatever the left side was (or right for RTL text), making it impossible to actually control the left and right padding independently on these elements.  This fixes that.


MacGritsch:
Thank you @NikkiDelRosso - had the same problem!

PauluzzNL:
Great patch. Please merge into master.

PR #81, opened 2018-01-29
Make first level of TOC bookmarks style not bold

abuzakham:


PauluzzNL:
In order to get such pull requests accepted it might help to add why this changes is a good idea / is an improvement.

nicolaasuni:
Thank for your contribution.

PR #83, opened 2018-02-13
each() is deprecated since PHP 7.2

svenpaulsen:
The each() function is deprecated in PHP 7.2. Calls of this function have been replaced with foreach() loops.


svenpaulsen:
I've noticed that this is a duplicate of #74.

PR #84, opened 2018-02-21
Restore PHP 5.3 compatibility Verified 11f0fd3

NiklasBr:
composer.json for TCPDF claims support for PHP 5.3 and short array syntax is not in this old version. This undos commit 865cd727fc29b080e8da1959566de5078a605c73 in char101:patch-1


troosan:
@nicolaasuni do you think you could release a new minor version with just this fix? BTW this PR does not « undo » the previous change, it just changes it to be PHP 5.3 compatible.

PR #85, opened 2018-02-28
Allow to set custom producer metadata

xelan:
This PR allows to override the default producer metadata in PDFs. For example, one may wish to remove the exact TCPDF version information, or to add application-specific suffixes in order to track the flow of generated PDF files.

The current default behavior is kept, i.e. the producer string `TCPDF <version> (http://www.tcpdf.org)` is added to all generated PDFs.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -512,6 +512,12 @@ class TCPDF {
 	protected $creator = '';
 
 	/**
+	 * Document producer.
+	 * @protected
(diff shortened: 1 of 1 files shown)


nicolaasuni:
Thank for your contribution.
Unfortunately this will not be merged here because the development of this branch is frozen and these changes are likely to create some backward compatibility issue.

PR #94, opened 2018-09-11
Security fix for disclosed stream wrapper attack

leofeyer:
This PR fixes a security vulnerability that has been presented at this years Black Hat USA 2018 conference. The white paper is available here:

https://github.com/s-n-t/presentations/blob/master/us-18-Thomas-It's-A-PHP-Unserialization-Vulnerability-Jim-But-Not-As-We-Know-It-wp.pdf

The fix checks if the file path begins with a stream wrapper other than `http` and `https` and throws an exception if so.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -6840,6 +6840,11 @@ class TCPDF {
 			// image from string
 			$imgdata = substr($file, 1);
 		} else { // image file
+			if ($file[0] === '*') {
+				// image as external stream
+				$file = substr($file, 1);
+				$exurl = $file;
+			}
 			$wrappers = stream_get_wrappers();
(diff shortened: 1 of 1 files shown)


leofeyer:
As pointed out by [Benni Mack](https://forge.typo3.org/users/320) on [Twitter](https://twitter.com/leofeyer/status/1040181223887433728), there is a generic [PHP Phar Stream Wrapper](https://github.com/TYPO3/phar-stream-wrapper) developed by the TYPO3 project, which is suitable to prevent the attack. Probably even better a solution than this fix.

leofeyer:
Thank you @nicolaasuni 👍

PR #97, opened 2018-09-21
Add @ operator to getimagesize function to mask the Warning error when the image is not found.

gaupoit:


nicolaasuni:
I am not sure what this PR is about.
Note that this project will only accept security fixes.

PR #99, opened 2018-09-24
Change file name to basename

anhtunguyen:


tcpdf.php
```diff
@@ -7558,6 +7558,8 @@ public function Output($name='doc.pdf', $dest='I') {
 		}
 		$dest = strtoupper($dest);
 		if ($dest[0] != 'F') {
+			$name = explode("/", $name);
```
NiklasBr: Why explode() the file name rather than pathinfo()? If explode is better, why explode on "/" rather than `DIRECTORY_SEPARATOR`?

tcpdf.php
```diff
@@ -7558,6 +7558,8 @@ public function Output($name='doc.pdf', $dest='I') {
...
+			$name = $name[count($name) - 1];
```
NiklasBr: Why only make sure that the basename should contain proper characters? Doesn't this create a problem and potential bug if anything but the basename contains unsupported characters?

tcpdf.php
staabm: I agree, that this should be `pathinfo` for x-platform compat

nicolaasuni:
Since this library is frozen I'll have to reject this PR. However I will consider a better approach for the new library.

staabm:
@nicolaasuni could you elaborate a bit? which is the 'new library'?

NiklasBr:
@nicolaasuni why?

PR #100, opened 2018-10-01
Fix unsupported operand types error when codepoints arrays are merged

jonasnutz:
I had some cases where $currentfont['subsetchars'] was 0, which lead to fatal errors because of unsupported operand types. This is fixed by this change.

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -2003,7 +2003,10 @@ class TCPDF_FONTS {
 			$chars = str_split($str);
 			$carr = array_map('ord', $chars);
 		}
-		$currentfont['subsetchars'] += array_fill_keys($carr, true);
+		if (is_array($currentfont['subsetchars']) && is_array($carr))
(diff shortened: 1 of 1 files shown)


include/tcpdf_fonts.php
```diff
@@ -2003,7 +2003,10 @@ public static function UTF8StringToArray($str, $isunicode=true, &$currentfont) {
 			$chars = str_split($str);
 			$carr = array_map('ord', $chars);
 		}
-		$currentfont['subsetchars'] += array_fill_keys($carr, true);
+		if (is_array($currentfont['subsetchars']) && is_array($carr))
+			$currentfont['subsetchars'] += array_fill_keys($carr, true);
+		else
+			array_merge($currentfont['subsetchars'], $carr);
```
staabm: the resulting array need to be assigned?

jonasnutz: Right, I thought array_merge does that automatically. Added a new commit with assignment.

PR #103, opened 2018-10-27
Fixing error on maxchwidth

j0k3r:
We got the error:
> Warning: max(): Array must contain at least one element

See https://github.com/wallabag/wallabag/issues/2776 & https://github.com/j0k3r/graby/pull/169

image

By that time we thought the project was dead so we don't provide a PR to fix that problem.
Instead we forked your repo and fixed the issue on our side https://github.com/wallabag/tcpdf
Instead of maintaining the fork, we think it's better to keep fixes in one place: your repo.

I recently rebased our repo with latest changes from your repo and it seems only that issue exists: https://github.com/tecnickcom/TCPDF/compare/master...wallabag:master


j0k3r:
That’s a good question.

j0k3r:
Looks like the project is still dead.

PR #104, opened 2018-11-14
Fix XRef parsing issue

seliquity:
TCPDF can't parse some files e.g http://samples.leanpub.com/serversforhackers-sample.pdf this change fixes that.


nicolaasuni:
This is deprecated, please use:
https://github.com/tecnickcom/tc-lib-pdf-parser

PR #108, opened 2019-01-04
Update custom_xmp line to comply with the RFC standard for XMP

gfgerard:
I believe the line where custom XMP is embedded into the rest of the XMP metadata is not correct.  

In the original any metadata embedded resulted in a non-standard XMP that didn't show up in e.g. Adobe Acrobat.


nicolaasuni:
The custom_xmp variable is empty by default, so you are probably setting some value?

defrance:
another way is to add a second RDF tags, it's what i have made and the solution is OK with factur-X validator

nicolaasuni:
I have added a new setExtraXMPRDF() function for that as the existing function has another purpose.

PR #111, opened 2019-01-17
Add arial fonts

krzysztofgwiazda:


nicolaasuni:
The Arial font is not free software and cannot be added.
You are free to add it in your installation if you have a license for it.

PR #114, opened 2019-02-13
Fix support for image URLs via zip wrapper

SailorMax:
After patch a5135e2cf02f8c935b71e9e7d676f0f1d31b1b9d you lost support images inside zip archives!

for example PHPSpreadsheet export to pdf can use html with image paths like: "zip://d:\\foo\\table.xlsx#xl/media/photo.png"

This patch fix this case.


nicolaasuni:
This library is currently frozen while working on the new one.
This change may pose some security risk so I will hold on that and review it for the new library.

PR #120, opened 2019-05-04
Allow to use multibyte string for file name

KohtaHorike:
Hello. 

To enable to use multibyte character for file name.  I would like to implement multibyte use option. 

Regards


nicolaasuni:
Unfortunately this library is frozen and I am not accepting any new feature at this time. However I will take this in consideration for the new version (tc-lib-pdf).

KohtaHorike:
@nicolaasuni Thanks for your reply! I am looking forward to new version :)

PR #126, opened 2019-07-25
Fix Array and string offset access syntax with curly braces is deprec…

remicollet:
…ated


remicollet:
Well...seems tehre is much more, are {} seems used neearly everywhere.....

PR #127, opened 2019-08-06
Let file_exists work with the "file://"-extension too

univietw:
Since "file://"-URLs are local files and basically the same as "/-" URLs they should be handled the same. Not doing this leads to errors for using HTML-Code in local instances.


nicolaasuni:
As I understand this poses a security issue. Since this library is frozen I will address this on the new library.

PR #128, opened 2019-08-09
Correct Base64 Bug in image url

LittleBigFox:
This patch allow to get base64 data in images url


nicolaasuni:
Missing CLA signature.
This PR cannot be accepted for security issues.
This shoul dbe resolved by the new tc-lib-pdf.

PR #129, opened 2019-08-14
Add new DroidSansFallback font for a better support of CJK characters

Molkobain:
Add new DroidSansFallback font for a better support of Chinese/Japanese/Korean characters


nicolaasuni:
This project is frozen.
Additional fonts can be added without the need to modify the core library.

PR #132, opened 2019-09-05
Fix bug with alphaPNG image and transaction

Adri74100:
Steps to reproduce :
- add an alphaPNG image
- start a transaction
- rollback the transaction
- add the same image a second time : it is not displayed

This is because the temporary files are deleted (in the destructor) when the $objcopy property is unsetted in rollbackTransaction().

````
$pdf = new TCPDF;

$pdf->AddPage();
$pdf->Image('logo.png');

$pdf->startTransaction();
$pdf->rollbackTransaction(true);

$pdf->AddPage();
$pdf->Image('logo.png');

$pdf->Output();
````

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -1818,6 +1818,12 @@ class TCPDF {
 	 */
 	protected $gdgammacache = array();
 
+	/**
+	 * If _destroyed() has already been called or not
+	 * @protected
+	 */
+	protected $destroyed = false;
+
 	//------------------------------------------------------------
 	// METHODS
 	//------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


eldy:
When the rollbackTransaction(true) is called,the method destroy is called.
And inside this method,there is loop to delete physically all images used. 

			if (isset($this->imagekeys)) {
				foreach($this->imagekeys as $file) {
					unlink($file);
				}
			}

So when you try to use it again later, the file does not exists anymore and your files is lost physically. This is acritical bug of last version.To fix thix, you can uncomment the line "unlink".

Adri74100:
@nicolaasuni May I ask why closing the request instead of merging the fix ?
I can improve the fix if necessary.

PR #133, opened 2019-09-11
Fix for PHP 7.4

crossjoin:
Fix for PHP 7.4 error because of wrong type for argument (i.e. an empty string): "chr() expects parameter 1 to be int, string given".
Occurred i.e. when calling TCPDF->__construct('P', 'mm', 'A4', true, 'UTF-8').


crossjoin:
Duplicate of #123

PR #135, opened 2019-09-19
Fix reported version

remicollet:


remicollet:
Please, also need to update the CHANGELOG file which is outdated.

PR #136, opened 2019-09-19
fix Undefined property: GLPIPDF::$imagekeys

remicollet:
```
  *** PHP Notice(8): Undefined property: GLPIPDF::$imagekeys
  *** PHP Warning(2): Invalid argument supplied for foreach()

```

P.S. GLPIPDF is a simple class extending TCDPF
https://github.com/glpi-project/glpi/blob/9.4/bugfixes/inc/glpipdf.class.php#L42

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -7796,8 +7796,10 @@ class TCPDF {
 				}
 				closedir($handle);
 			}
-			foreach($this->imagekeys as $file) {
-				unlink($file);
+			if (isset($this->imagekeys)) {
+				foreach($this->imagekeys as $file) {
+					unlink($file);
+				}
 			}
 		}
 		$preserve = array(


remicollet:
Introduce in b41a5ca3114c18214ec2871fb68923e6d7bb9e92

remicollet:
This happen because __destroy is call various time
Another solution will be to` $this->imagekeys = array(); `after the unset

eclipxe13:
As `_destroy` method is called on `__destruct` and also registered using `register_shutdown_function` (on `__construct`) then variables for `$this` should be checked with `isset`.

Another question is, why call `register_shutdown_function` with the same callable as in `__destruct`? Is this a compatibility issue for first implementations of PHP Objects? Does it really create value? Maybe if we let the garbage collector do its job it will destroy the object on time and don't need to execute twice `_destroy`...

Please, merge this and release 6.3.1, please.

eclipxe13:
> 
> Another question is, why call `register_shutdown_function` with the same callable as in `__destruct`? Is this a compatibility issue for first implementations of PHP Objects? Does it really create value? Maybe if we let the garbage collector do its job it will destroy the object on time and don't need to execute twice `_destroy`...
> 

I see, looks like `__destruct` is not called in case of fatal errors, so, there is no other way to call `_destroy` on that scenario.

The case mentioned on comment `Call cleanup method after script execution finishes or exit() is called.` is false according to documentation. It should be `Call cleanup method after script execution finishes on fatal error.`

remicollet:
Thanks.

remicollet:
Please apply this one again.
6.3.3 is broken

williamdes:
cc @nicolaasuni 

Also please keep me updated about releases because I am responsible for the Debian packaging of tcpdf.

remicollet:
In fact, simply revert 7dbdfa405123c2fbca4864d87a143f7977403767

nicolaasuni:
It should be fixed now. I have mixed up some PRs...

nicolaasuni:
@remicollet thank you!
I am hoping to switch to the new tc-lib-pdf at some point and stop updating this.

PR #137, opened 2019-09-20
PHP 7.4 - cannot use array offset on integers

fisharebest:
PHP 7.3 casts integers to strings.
PHP 7.4 throws the error `Trying to access array offset on value of type int`


fisharebest:
PHP 7.4 will be released in 2 days.  Can we get this merged before then?

DevinWalker:
Any progress on this?

fisharebest:
@DevinWalker - FYI, I am currently using [composer-patches](https://github.com/cweagans/composer-patches) to apply this (and other) patches locally.

DevinWalker:
Thanks @fisharebest that's helpful. I'll proceed with that until it's released.

PR #140, opened 2019-09-26
Possibility to set text in Footer

wincelau:
Adding option in setFooterData Method to let the possibility to add text in footer


nicolaasuni:
Closing this as I am not accepting new feature for this library.
New features should go in tc-lib pdf.

PR #142, opened 2019-09-30
Prevents crash and files deletion.

DominiqueFERET:
Added an @ in front of unlink to avoid the crash.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -7791,14 +7791,14 @@ class TCPDF {
 			if ($handle = opendir(K_PATH_CACHE)) {
 				while ( false !== ( $file_name = readdir( $handle ) ) ) {
 					if (strpos($file_name, '__tcpdf_'.$this->file_id.'_') === 0) {
-						@unlink(K_PATH_CACHE.$file_name);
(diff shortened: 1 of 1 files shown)


DominiqueFERET:
Yes, it's true, but, in this case, if(is_file($file)) unlink($file); is better.

dfontenot:
Running into this issue where http:// links to images are ending up in `$this->imagekeys` causing unlink() to crash. Needing something like this MR to fix the issue.

DominiqueFERET:
> 
> 
> Running into this issue where http:// links to images are ending up in `$this->imagekeys` causing unlink() to crash. Needing something like this MR to fix the issue.

Thank you for this correction. You're right, it's better.

it-can:
Also having problems with this, can anyone merge this and release a new version?

it-can:
@nicolaasuni can you merge this ?

doits:
:+1: this also prevent deleting files not generated by tcpdf! Otherwise it ate (=deleted) one of my .png each time a pdf was generated, this is a serious bug leading to data loss!

NiklasBr:
ping @nicolaasuni

PR #143, opened 2019-09-30
Prevent crash in case of no list access in cache path

DaRealFreak:
added @ to prevent crash in case that we have no list access in the cache path


it-can:
@nicolaasuni can you merge this?

PR #134, opened 2019-09-13
Fix Array and string offset access syntax with curly braces is deprec…

remicollet:
…ated

For upcoming PHP 7.4

PR #144, opened 2019-10-20
PHP 7.4 - some more curly braces to fix

stronk7:
While #134 fixed the majority of, now deprecated, curly braces
it seems that a few were left out.

`ag '(\$[0-9a-zA-Z_]+|\))(\[[\w$]+\])?{[^\n=}]+}' --php | sort | uniq`

This commit just converts them to square braces.

### DIFF
--- a/include/barcodes/datamatrix.php
+++ b/include/barcodes/datamatrix.php
@@ -629,7 +629,7 @@ class Datamatrix {
 					if ($numch[ENC_C40] == $numch[ENC_X12]) {
 						$k = ($pos + $charscount + 1);
 						while ($k < $data_length) {
-							$tmpchr = ord($data{$k});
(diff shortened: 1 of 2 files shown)


stronk7:
I'm modifying this under the terms of the GNU LESSER GENERAL PUBLIC LICENSE Version 3, 29 June 2007 and providing the modifications to you under those very same terms.

So I'm not going to sign absolutely anything, I'm afraid. Feel free to ignore this pull request, pick it, change its authorship to another, signed, contributor or whatever you prefer.

Ciao :-)

PS: Thanks for your awesome library!

fisharebest:
@stronk7 - If you publish your code under the GPL, it makes it **impossible** to use in a project with a more restrictive licence.

Suggestion: place the code is in the public domain - and then someone can add it under a CLA.

PR #145, opened 2019-10-20
Problem with $imagekeys undefined or unlinked

woytam:
PR #136 removed problem with undefined `$imagekeys`, but the solution skipped cleaning files in `$imagekeys`.
The problem about this is that `_destroy` function is called 3 times (from `Close` function, as `shutdown_handler` and from `__desctruct` function). Only first time is called as `destroyall == false`, but unset `$imagekeys`. My solution is to preserve `$imagekeys` to be cleaned by call with `$imagekeys == true`.


nicolaasuni:
I have just pushed another release. It is now deleting only the images in the cache folder if they exist.

PR #146, opened 2019-10-21
FIX SVGPath elliptical arc with rx/ry=0 + z should return to initial …

baptistepillot:
Two fixes proposals into this PR :

- For a path like this one, which draw a simple "A" : 
```m 2594.06,816.219 h 318.11 l -159.34,302.461 z m -214.13,-220.84 322.58,590.951 h 105.67 l 328.19,-590.951 h -108.46 l -75.49,145.363 h -399.74 l -75.47,-145.363```
When the "z" command comes, the path closes and the coordinates should return back to the initial point coordinates 2594.06,816.219. See specs https://www.w3.org/TR/SVG/paths.html#PathDataMovetoCommands and https://www.w3.org/TR/SVG/paths.html#PathDataClosePathCommand.
I added $xinitial and $yinitial to keep in memory this "initial point", and return back coordinates to it on "z". It worked for this case, please check if this causes no problem with mines/maxes.

Here is a link to the SVG image I embed into the TCPDF generator using imageSVG() :
http://next.pillot.fr/middle-z.svg
Into your browser : works well
Into a PDF : looks bad
Into a PDF, with my proposal : looks good

- I had a "warning division by zero" on some cases where softwares create 0 radiuses. A little round of it avoids these alerts (max(..., .000000001)).

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -23406,7 +23406,7 @@ Putting 1 is equivalent to putting 0 and calling Ln() just after. Default value:
 	 */
 	protected function SVGPath($d, $style='') {
 		if ($this->state != 2) {
-			 return;
+			return;
 		}
 		// set fill/stroke style
 		$op = TCPDF_STATIC::getPathPaintOperator($style, '');
@@ -23426,6 +23426,8 @@ Putting 1 is equivalent to putting 0 and calling Ln() just after. Default value:
 		$xmax = 0;
 		$ymin = 2147483647;
 		$ymax = 0;
+		$xinitial = 0;
+		$yinitial = 0;
 		$relcoord = false;
 		$minlen = (0.01 / $this->k); // minimum acceptable length (3 point)
 		$firstcmd = true; // used to print first point
@@ -23470,6 +23472,8 @@ Putting 1 is equivalent to putting 0 and calling Ln() just after. Default value:
 								if ($ck == 1) {
 									$this->_outPoint($x, $y);
 									$firstcmd = false;
+									$xinitial = $x;
+									$yinitial = $y;
 								} else {
 									$this->_outLine($x, $y);
 								}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Cyrille37** [comment]
Hello
Thanks a lot for this correction !

I have a bug with 'c' command (curveto) that is solved by your 'm' command patch.

The right result (with Inkscape, Illustrator, eye of Mate, ...) with your patch :
image

The bad result without the patch:
image

The SVG is : 
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
   xmlns:dc="http://purl.org/dc/elements/1.1/"
   xmlns:cc="http://creativecommons.org/ns#"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg"
   id="svg1181"
   version="1.1"
   viewBox="0 0 52 74"
   height="74mm"
   width="52mm">
  <defs
     id="defs1175" />
  <metadata
     id="metadata1178">
    <rdf:RDF>
      <cc:Work
         rdf:about="">
        <dc:format>image/svg+xml</dc:format>
        <dc:type
           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
        <dc:title></dc:title>
      </cc:Work>
    </rdf:RDF>
  </metadata>
  <g
     transform="translate(0,-223)"
     id="layer1">
    <path
       clip-path="none"
       mask="none"
       d="m 6.7730753,227.66821 c -0.225772,-0.7588 -0.789497,-1.1592 -1.466813,-1.1592 -0.677317,0 -1.230811,0.42086 -1.405785,1.1592 z m 0.984931,1.91836 c -0.595121,0.74928 -1.415662,1.20047 -2.308167,1.21035 -1.73386,0 -2.739251,-1.30242 -2.739251,-2.61543 0,-1.31336 1.005391,-2.61613 2.739251,-2.61613 1.036082,0 2.57486,0.85158 2.57486,2.7904 v 0.19473 h -4.124222 c 0.06173,0.75916 0.769742,1.30277 1.518318,1.30277 0.574661,0 1.067126,-0.27692 1.436122,-0.76938 z"
       style="fill:#5ba7dd;stroke-width:0.35276896"
       id="path38" />
  </g>
</svg>
```

nicolaasuni:
@baptistepillot Would yu please check this also in https://github.com/tecnickcom/tc-lib-pdf-graph ?

Cyrille37:
Thanks for merge :-)

baptistepillot:
> @baptistepillot Would yu please check this also in https://github.com/tecnickcom/tc-lib-pdf-graph ?

I will try (but I do not use this library).

Thanks for merge, I am happy I helped this great project, and proud to be a (very little) part of it.

PR #147, opened 2019-10-23
Changes PageBreakTrigger in footer in order to make fitBlock work

maechler:
When we are in the footer we should set `$this->PageBreakTrigger` to the page height in order to make `fitBlock` work. This is e.g. an issue if you want to render an image in the footer with the option `fitonpage` being set.


nicolaasuni:
I would like to merge this but I am not confident is going to break some existing implementation.
As a workaround you can apply the patch via Composer.
In the meanwhile please chek if you can contribute to tc-lib-pdf

PR #151, opened 2019-11-27
Suggestion for vertical alignment of images in HTML cells

gianlucagiacometti:
I tested it with two inline images and some text. One image with the attribute height specified, the other one without it.
Seems to work for both top and middle values.


nicolaasuni:
Thank you for your contributiion.
I am not accepting new features here as I prefer to concentrate on tc-lib-pdf project.
Additionally the hardcoded font ratio are too simplistic for a generic library and they will break in a lot of cases.

PR #164, opened 2020-01-15
Fixed datamatrix.php for PHP 7.4

mhedderich:
Resolved deprecation error in datamatrix.php: "Deprecated: Array and string offset access syntax with curly braces is deprecated"

### DIFF
--- a/include/barcodes/datamatrix.php
+++ b/include/barcodes/datamatrix.php
@@ -629,7 +629,7 @@ class Datamatrix {
 					if ($numch[ENC_C40] == $numch[ENC_X12]) {
 						$k = ($pos + $charscount + 1);
 						while ($k < $data_length) {
-							$tmpchr = ord($data{$k});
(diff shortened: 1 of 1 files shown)


nicolaasuni:
Please note that is better to use tc-lib-barcode that is compatible with TCPDF.
The barcodes in TCPDF are bugy and unmaintained.

include/barcodes/datamatrix.php
```diff
@@ -629,7 +629,7 @@ protected function lookAheadTest($data, $pos, $mode) {
 					if ($numch[ENC_C40] == $numch[ENC_X12]) {
 						$k = ($pos + $charscount + 1);
 						while ($k < $data_length) {
-							$tmpchr = ord($data{$k});
+							$tmpchr = ord($data[$k]);
```
anaselbahrawy, line 632: yes it's solved in PHP version 7.4

PR #172, opened 2020-02-03
Allow to use AWS S3 through a stream prefixer

reindert-vetter:
The problem:
`s3://alwins-test-bucket/app/labels/test.pdf` is not seen as a local disk to which you can write. Now I get the following error; `Unable to create output file: s3://alwins-test-bucket/app/labels/test.pdf`. That while streaming to S3 can be treated as a local disk.


nicolaasuni:
I am not merging this because of some security issues.
You should be able to patch the current TCPDF version using Composer.
Hopefully this will be available in tc-lib-pdf at some point

PR #174, opened 2020-02-07
Do not unlink original file !

upsite4k:
this function deletes the original file not a copy or a reference !
$this->imagekeys holds the complete path and filename of the file which should be embed, so its not advised to delete this  :)


nicolaasuni:
This should be solved by a previously merged PR.
Please let me know if this is not the case.

upsite4k:
you fixed it in commit 510070b5e068e10d97b8cb0e25295e54755ec09d

thanks

Issue #179, opened 2020-02-13
In TCPDF 6.3.4 after pdf render images with base64 not appear


eliel-jr:
I'm use PHP 7.2.5 and after update TCPDF to 6.3.4, not rendering images with base64

images with filepath render OK.

on v6.2.13 rendering OK.

HTML example in text that not rendering base64

[html_example.txt](https://github.com/tecnickcom/TCPDF/files/4198078/html_example.txt)

eliel-jr:
I'm testing versions: v6.3.4, v6.3.3, v6.3.2, v6.3.1, v6.3.0, v6.2.26, v6.2.25, v6.2.23, v6.2.22 and all broken

in v6.2.21 render OK.

LittleBigFox:
Hi!

I'd the same problem, just replace the tcpdf/include/tcpdf_static.php function file_exists (L1879 in tcpdf 6.3.4) like this:

`

public static function file_exists($filename) {
    if (preg_match('|^https?://|', $filename) == 1) {
        return self::url_exists($filename);
    } 

    if(stripos($filename,'data:')===0) {
        return true;
    }

    if (strpos($filename, '://')) {
          return false; // only support http and https wrappers for security reasons
    }
    
    return @file_exists($filename);
}
`

Vitexus:
https://github.com/spipu/html2pdf/issues/413

clickmeplz:
Version 6.2.21 works fine in local and server. I have 6.3.2 and render Ok in local and not works in server.
Above solution not working for me. I checked with 6.3.2 and 6.4.4

detook:
Do this
`<img src="@/9j/4AAQ....`
instead of
`<img src="data:image/png;base64,/9j/4AAQ....`

compojoom:
@detook solution works, but so is @LittleBigFox's if one is ok with modifying the static file. 

I personally find it very sad that we have to add an @ instead of the default: 
```
data:image/png;base64
```

I've created a PR with @LittleBigFox's fix:
https://github.com/tecnickcom/TCPDF/pull/671

nicolaasuni:
[...]

PR #180, opened 2020-02-13
[FIX] If filename is base64 data:, file exists = true

eliel-jr:
FIX ISSUE [#179 ](https://github.com/tecnickcom/TCPDF/issues/179)


nicolaasuni:
I can't merge this for security reasons.
You can apply this path via Composer.

eliel-jr:
OK.

please, fix other solution to load base64 images.

i need

carlos-granados:
Why is this a security issue? I don't see the security problem anywhere. I think this should be accepted as right now it is impossible to use base64 images in pdf files

PR #187, opened 2020-02-22
php7.3 and imagemask fatal error fix

vavrecan:
fixes

PHP message: PHP Fatal error:  Uncaught ErrorException: unlink(/tmp/__tcpdf_351ebd7891d67ccef6584da0eee7e6f5_imgmask_alpha_3fd2ef0623f637f931694c0b8e4c33df)

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -7787,7 +7787,7 @@ class TCPDF {
 			}
 			if (isset($this->imagekeys)) {
 				foreach($this->imagekeys as $file) {
-					if (strpos($file, K_PATH_CACHE) === 0) {
+					if (strpos($file, K_PATH_CACHE) === 0 && file_exists($file)) {
 						@unlink($file);
 					}
 				}


xelan:
We stumbled across the same issue here... Due to the fatal error, the PDF is created but other shutdown handlers etc. are not executed correctly after that.

By the way, wouldn't `is_writable` be better than `file_exists`, as a write operation is attempted? According to the [PHP docs](https://www.php.net/manual/en/function.is-writable), `is_writable` returns _true_ if the filename exists and is writable.

vavrecan:
> We stumbled across the same issue here... Due to the fatal error, the PDF is created but other shutdown handlers etc. are not executed correctly after that.
> 
> By the way, wouldn't `is_writable` be better than `file_exists`, as a write operation is attempted? According to the [PHP docs](https://www.php.net/manual/en/function.is-writable), `is_writable` returns _true_ if the filename exists and is writable.

the file does not exists in /tmp directory at all. Not sure where it was created at the first place (in current code they just ignore errors using old at-before-function used in older php version)

xelan:
> the file does not exists in /tmp directory at all. Not sure where it was created at the first place (in current code they just ignore errors using old at-before-function used in older php version)

Thanks, I was wondering the same, but thought that there might be cases where those temp files are used... Apparently the K_PATH_CACHE constant [defaults](https://github.com/tecnickcom/TCPDF/blob/6.3.5/tcpdf_autoconfig.php#L119) to the ini setting `upload_tmp_dir` if set, or `sys_get_temp_dir()` otherwise...

PR #194, opened 2020-03-06
ADD: check for file really exist before delete, if not i got an error…

alfreddagenais:
## Problem this Pull Request solves
Check for file really exist before delete, if not i got an error No such file or directory

## How has this been tested
I tested directly online with this library

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -6903,7 +6903,9 @@ class TCPDF {
 			unset($imgdata);
 			$imsize = @getimagesize($file);
 			if ($imsize === FALSE) {
-				unlink($file);
+				if( file_exists($file) ){
+					@unlink($file);
+				}
 				$file = $original_file;
 			}
 		}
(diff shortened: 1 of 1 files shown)


SunMar:
https://github.com/tecnickcom/TCPDF/pull/194/files#diff-35e08735787098d1dc471c0d930ffb71L7783 does not look like it needs a `file_exists()`, since it's part of a directory scan. You should only add a `file_exists()` before `unlink()` if there is a valid use case for the file being missing. If the use case means the file should always exist, adding a `file_exists()` is only going to hide the real problem, making it much harder to find.

Do you have any information on what error/bug/issue you are actually trying to solve?

alfreddagenais:
> https://github.com/tecnickcom/TCPDF/pull/194/files#diff-35e08735787098d1dc471c0d930ffb71L7783 does not look like it needs a `file_exists()`, since it's part of a directory scan. You should only add a `file_exists()` before `unlink()` if there is a valid use case for the file being missing. If the use case means the file should always exist, adding a `file_exists()` is only going to hide the real problem, making it much harder to find.
> 
> Do you have any information on what error/bug/issue you are actually trying to solve?

Yes you have right! I just saw a `unlink()` method at this place and I added `file_exists()` just be sure if they have no more error. Normally, we don't need that part of the code. The only part I suggest it's keep the   `@` with `@unlink()` to help when it's not possible to delete the file. What do you think?

nicolaasuni:
Unfortunately the file_exists() functions has a known vulnerability.
Adding "@" before unlink() should be a safer option.

alfreddagenais:
> Unfortunately the file_exists() functions has a known vulnerability.
> Adding "@" before unlink() should be a safer option.

Ohh nice! I did not know that. Thanks for the input.

PR #206, opened 2020-04-10
Remove file_id from the cloned object before destruction

### LINKED ISSUE #205: Usage of rollbackTransaction() removes all files from cache including previously copied images
The usage of `startTransaction()`/`rollbackTransaction()` is a recommended way to figuring out the precise height of the element before actual rendering. But it has this side effect, that it removes all the data from cache directory, including the copied images. It is well visible with header images if they are resized while printing.

The issue can be reproduced with the following code (based on example 047):
https://gist.github.com/Jakuje/d88c837f6520416964926fac5187ccbf

The first image is visible, but the second one is not. This worked fine with some old version of TCPDF (before the image file caching was introduced?), but does not work now.

I spent some time figuring out where the issue lies and it turns out that it is a combination of several aspects:
 * the `rollbackTransaction()` needs to be called with `self` argument
 * the image needs to be resized
 * ...

What is going on here is that the image is copied to the tmp directory in the first object, which is cloned, the `rollbackTransaction()` calls `_destroy(true, true)` with correct arguments to preserve this information. But just few lines below, it is called again, this time as `_destroy(true)` from destructor, which leads to the erasing of the `tmp` directory.

This was bothering me for quite some time and I believe the correct fix will be to have a boolean property which should be set in `_destroy()` to prevent the other call to `_destroy()` in this case. More about the proposed solution will come soon in pull request.

Jakuje:
The `_destroy()` function called from destructor is destructive and touches the same files as the original object is still using when cloned through `start/commit/rollbackTransaction()`. The current code does not handle this corner case and leaves this function to be called by function destructor, which is taking the `/tmp/` files with it.

The proposed patch modifies the functions handling transaction to remove `file_id` from the objects before the final `_destroy` is called from destructor. 

Fixes #205

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -7774,7 +7774,7 @@ class TCPDF {
 		if (isset(self::$cleaned_ids[$this->file_id])) {
 			$destroyall = false;
 		}
-		if ($destroyall AND !$preserve_objcopy) {
+		if ($destroyall AND !$preserve_objcopy && isset($this->file_id)) {
 			self::$cleaned_ids[$this->file_id] = true;
 			// remove all temporary files
 			if ($handle = @opendir(K_PATH_CACHE)) {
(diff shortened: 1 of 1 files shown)


Eldofas:
I also came across this one. Thanks Jakuje for the pull request.

Jakuje:
@nicolaasuni would you mind commenting on why this was closed without merging the change?

Issue #230, opened 2020-05-15
mb_internal_encoding is set to ASCII


grizzm0:
In `TCPDF::__construct` the internal encoding is set to ASCII and changed back to whatever it was before during destruct.

However, if you use any mb_ function outside of TCPDF after a TCPDF instance has been created all your mb_ functions will use ASCII as encoding.

We're using UTF8 (which is PHP default) as mb_internal_encoding. After creating an instance of TCPDF we then used mb_substr to truncate some long strings in our PDFs. However, as the internal encoding now was set to ASCII this sometimes failed as our strings contains UTF8.

I've been looking through the code and can't seem to find any good reason to why the internal encoding is changed to ASCII.

bkonetzny:
Same issue here, this was really hard to debug why the mb_ functions behaved different. The PR seems to provide a valid fix for this.

LinkingYou:
Is there a reason why the internal encoding is set to ASCII at all? 

I am using Dompdf and get problems with line breaks because umlauts seem to calculate the wrong line length. If I comment out the passage `mb_internal_encoding('ASCII');` , everything works as expected.

nicolaasuni:
Thank you for reporting this. TCPDF 6 is now a legacy project and is officially **deprecated**, so no further fixes or features will be merged into this repository. As part of this transition, we are closing older issues. Please migrate to **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** and, if the same problem still occurs there, open a new issue with updated reproduction steps. For more context, see the pinned announcement: [Thank You - and the Future of TCPDF](https://github.com/tecnickcom/TCPDF/issues/867).

PR #231, opened 2020-05-20
Update tcpdf.php

ollea:
Solution to issue: https://github.com/tecnickcom/TCPDF/issues/230


Daeroni:
Should the mb_internal_encoding be restored at the end of `getHtmlDomArray` (and the class variable `internal_encoding` coverted to a local variable)? There are two calls to the method, in `writeHTML` and in `addHTMLTOC`, the latter does not restore the encoding.

PR #237, opened 2020-06-14
Fix "undefined variable" if "helvetica" font not loaded in tcpdf.php

lubosdz:
On line [line 9863](https://github.com/tecnickcom/TCPDF/blob/master/tcpdf.php#L9863) in tcpdf.php the variable `$font` may contain also `false` if helvetica font not loaded, which throws error.


interduo:
When this willbe merged?

lubosdz:
@nicolaasuni PR closed - any reasoning? Error is still valid. Or is it fixed somewhere else .. ?

PR #243, opened 2020-06-22
TSA timestamp signature feature

francescouri-coretech:
Solution for not implemented function to add a timestamp signature using simple libraries phpcmstimestamper and phpseclib

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,9 +22,13 @@
     }
   ],
   "require": {
-    "php": ">=5.3.0"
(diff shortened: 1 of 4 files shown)


nicolaasuni:

JMHAVANCE:
@francescouri-coretech I'm trying to implement your solution but my PDF's signature is invalid after applying the TSA Timestamp.

Is there any way I can reach out to you for help? (disarmyouwitha@gmail.com)

PR #244, opened 2020-06-30
PHP8 errors and several other fixes

brainfoolong:
I'm testing PHP8 with our apps that use TCPDF. There are things in TCPDF that are incompatible with PHP8 and throw errors. I've took the time to fix the issues and also fixed other issues that i came across.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.php text eol=lf
+/.github export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.php eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-*.php eol=lf
+*.php text eol=lf
--- a/include/tcpdf_static.php
+++ b/include/tcpdf_static.php
@@ -276,7 +276,7 @@ class TCPDF_STATIC {
 	/**
 	 * Determine whether a string is empty.
 	 * @param $str (string) string to be checked
-	 * @return boolean true if string is empty
+	 * @return bool true if string is empty
 	 * @since 4.5.044 (2009-04-16)
 	 * @public static
 	 */
--- a/include/tcpdf_static.php
+++ b/include/tcpdf_static.php
@@ -1859,7 +1859,7 @@ class TCPDF_STATIC {
 	public static function encodeUrlQuery($url) {
 		$urlData = parse_url($url);
 		if (isset($urlData['query']) && $urlData['query']) {
-			$urlQueryData = [];
+			$urlQueryData = array();
 			parse_str(urldecode($urlData['query']), $urlQueryData);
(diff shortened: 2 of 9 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.php text eol=lf
```
williamdes, line 1: Could you export ignore `.github` ?

include/tcpdf_static.php
```diff
@@ -1859,7 +1859,7 @@ public static function url_exists($url) {
 	public static function encodeUrlQuery($url) {
 		$urlData = parse_url($url);
 		if (isset($urlData['query']) && $urlData['query']) {
-			$urlQueryData = array();
+			$urlQueryData = [];
```
williamdes, line 1862: You should revert this change to keep maximal php versions coverage

.gitattributes
brainfoolong: Hi, what do mean with "export"?

tcpdf.php
```diff
@@ -4291,6 +4291,7 @@ public function AddFont($family, $style='', $fontfile='', $subset='default') {
 				$fontfile = TCPDF_FONTS::getFontFullPath($tmp_fontfile, $fontdir);
 			}
 		}
+		$originalsize = $size1 = $size2 = null;
```
williamdes, line 4294:
```suggestion
		$originalsize = null;
		$size1 = null;
		$size2 = null;
```

include/tcpdf_static.php
brainfoolong: I saw that in several other parts of the code that are already short array syntax is used, so i've changed it there too.

brainfoolong: Ah, nevermind. I was wrong, i'll revert.

.gitattributes
williamdes: Hi
The file controls `git archive` and basically, excluding some files nobody wants on their production servers is always cool. In debian we do not need the github folder for example 

https://github.com/williamdes/mariadb-mysql-kbs/blob/main/.gitattributes#L15
https://salsa.debian.org/phpmyadmin-team/mariadb-mysql-kbs/-/tree/upstream

Tcpdf: https://salsa.debian.org/phpmyadmin-team/tcpdf/-/tree/upstream

brainfoolong: Thx. I've done that.

williamdes: Thank you!
One note about your commit message it is more "do not export github folder in vendor tarballs" ;)

brainfoolong: Thanks for the info.

mircobabini:
I know that you want to keep doing the great job your are doing but... when will you push this hotfix? It's causing us a lot of issues. :) Thanks.

williamdes:
@nicolaasuni is this PR rejected or merged in some way?

williamdes:
@nicolaasuni Was the PR closed by mistake ?

nicolaasuni:

PR #245, opened 2020-07-01
Fix 'Constant xxx already defined'

suthio:
I'm experiencing the following problems in my environment.

```
Constant K_BLANK_IMAGE already defined
```

I checked and found that the error is caused by the following two libraries.

https://github.com/tecnickcom/TCPDF
https://github.com/DALTCORE/lara-pdf-merger

The same constants are defined.

https://github.com/tecnickcom/TCPDF/blob/master/config/tcpdf_config.php#L87
https://github.com/DALTCORE/lara-pdf-merger/blob/2.0/src/LynX39/LaraPdfMerger/tcpdf/tcpdf_autoconfig.php#L128

Fixed to work when the same constant is defined multiple times.

I hope you can confirm this.

@nicolaasuni

### DIFF
--- a/config/tcpdf_config.php
+++ b/config/tcpdf_config.php
@@ -84,143 +84,198 @@
 /**
  * Generic name for a blank image.
  */
-define ('K_BLANK_IMAGE', '_blank.png');
+if (!defined('K_BLANK_IMAGE')) {
+    define('K_BLANK_IMAGE', '_blank.png');
+}
 
 /**
  * Page format.
  */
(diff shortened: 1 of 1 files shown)


suthio:
@nicolaasuni Why did you close this?

nicolaasuni:
This is a specific implementation issue.
TCPDF is feature freeze in favour of tc-lib-pdf
Why not trying to contribute to tc-lib-pdf instead?

nicolaasuni:

PR #250, opened 2020-07-28
allow styles on <HR> tags

Chrissi2812:
Allows to pass some styles to `<hr>` tags

**Example:**
```html
<hr style="color: #ff0000; dash: 4">
```
**Render:**
image

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -18865,7 +18865,16 @@ Putting 1 is equivalent to putting 0 and calling Ln() just after. Default value:
 				}
 				$prevlinewidth = $this->GetLineWidth();
 				$this->SetLineWidth($hrHeight);
-				$this->Line($x, $y, $x + $hrWidth, $y);
+
+				$lineStyle = array(
+					'color' => $tag['fgcolor'],
+					'cap'   => $tag['style']['cap'],
+					'join'  => $tag['style']['join'],
+					'dash'  => $tag['style']['dash'],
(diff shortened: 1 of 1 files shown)


nicolaasuni (changes requested):
These new tag keys are not defined anywhere.
You should add at least some default values.

Chrissi2812:
The styles are now filtered before being added, so the default values will be used.

nicolaasuni (approved):

fzieris:
> The styles are now filtered before being added, so the default values will be used.

No, they are not (see issue #347). Why use these four non-standard CSS property names anyway?

vlakoff:
The OP example shows a CSS attribute `color`, whereas the code implements an HTML attribute `fgcolor`. Which one is the proper?

vlakoff:
Nevermind, I found the following in the code:

```
$dom[$key]['fgcolor'] = TCPDF_COLORS::convertHTMLColorToDec($dom[$key]['style']['color'], $this->spot_colors);
```

… so the `color` CSS attribute is passed to that `fgcolor` property.

trasher:
This change seems to cause several issues. First of all, undiefined index are still occuring; the only way to fix on my side is to declare empty css rules.... Not very convenient.

But I also have another problem: many times when using an HR tag (even not styled), the output PDF texts becames "blank" (white text on white page...). This issue is difficult to reproduce; I can get it fixed removing present PR changes :/

PR #263, opened 2020-08-18
temporary files never deleted

feddar:
When creating PNG files, temporary file are not removed (search for __tcpdf_* in your temp directory). After this changes temp directory never has temporary files created by tcpdf


nicolaasuni (approved):

nicolaasuni:
I had to revert this after merging because if I remember correctly these temporary files are removed by the _destroy() function.
We keep these files in the cache to be reused multiple times.

PR #268, opened 2020-08-25
Fix #267 -  Empty result for big reports with writeHTML()

### LINKED ISSUE #267: Empty result for big reports with writeHTML()
The error was noticed 10 years ago. When trying to generate a PDF from a large HTML file, an empty file is generated, and no error messages are generated.

The report generation dies because of a preg_replace() used by TCPDF. It was a PHP bug for older versions (mine is 5.3.3), preg_replace() returns null silently when the string is too long. See this:

https://sourceforge.net/p/tcpdf/discussion/435311/thread/f7f96b04/
https://stackoverflow.com/questions/14870397/tcpdf-displays-blank-pdf-if-query-result-is-big

eisberg:
Fix issue #267

Optimized regular expressions, which fixed a bug with large HTML

Test code:

```php
 <?php
require_once __DIR__ . './vendor/tecnickcom/tcpdf/tcpdf.php';
$html = file_get_contents('./test.html');

$pdf = new TCPDF('L', PDF_UNIT, 'A2', true, 'UTF-8', false);

$pdf->AddPage();
$pdf->SetFont('helvetica', null, 12);

// output the HTML content
$pdf->writeHTML($html, true, false, true, false, '');
$pdf->Output('example.pdf', 'I');
```

test.html in zip archive:
[test.html.zip](https://github.com/tecnickcom/TCPDF/files/5123493/test.html.zip)

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -16354,7 +16354,7 @@ class TCPDF {
 		// create a special tag to contain the CSS array (used for table content)
 		$csstagarray = '<cssarray>'.htmlentities(json_encode($css)).'</cssarray>';
 		// remove head and style blocks
-		$html = preg_replace('/<head([^\>]*)>(.*?)<\/head>/siU', '', $html);
+		$html = preg_replace('/<head([^\>]*)>(.*)?<\/head>/siU', '', $html);
 		$html = preg_replace('/<style([^\>]*)>([^\<]*)<\/style>/isU', '', $html);
 		// define block tags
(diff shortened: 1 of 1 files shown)


skillbox-koutja:
Mainteners, apply please

eisberg:
any news?

NiklasBr:
ping @nicolaasuni :)

freddieleeman:
I can confirm that this fix works but I am unable to work out why. Why does the current regex fail on large strings, and not small ones (<1000095 characters)? The threshold appears to be at 1000094 characters, any more than that and the regex fails.

The bug can also be fixed by just removing the '/U' quantifier.

eisberg:
The version with this patch was tested in production, no problems were found

nicolaasuni (approved):

PR #300, opened 2020-11-19
Fix CS to PSR-12 standard

mvorisek:
the only changed files are:
- `.gitignore`
- `.php_cs.dist`
- `composer.json`

the php files are unmodified and strictly only fixed by CS Fixer

to prevent conflicts, this PR should be merged immediatelly

PS: feel free to rerun CS Fixer, but must be run twice to stabilize and each run takes almost a day (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5279)


.php_cs.dist
```diff
@@ -0,0 +1,75 @@
+<?php
+
+$finder = PhpCsFixer\Finder::create()
```
keradus, line 3: pssst, you may disable autogenerated, 1MB+ files that are too slow to be processed

.php_cs.dist
```diff
@@ -0,0 +1,75 @@
...
+        'yoda_style' => [
+            'equal' => false,
+            'identical' => false,
+        ],
+        'native_function_invocation' => false,
+        'void_return' => false,
+        'blank_line_before_statement' => [
+            'statements' => ['break', 'continue', 'declare', 'return', 'throw', 'exit'],
+        ],
+        'combine_consecutive_issets' => false,
```
keradus, line 35: if there is a need to disable 30 rules, maybe it's better to explicitly enable full rulesets like PSR2 and maybe few rules on top of it?

mvorisek: @keradus I have simple config - https://github.com/mvorisek/TCPDF/tree/psr12_cs_extended - as far I as remember there is one rule that does not finish at all. I belive, in CS fixer, there are several small issues with extreme/exponencial complexity. That is why I opened https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5279

mvorisek: @keradus the issue is with `array_indentation`, when not disabled, the CS Fixer never finish (I waited more than 24 hours)

keradus: https://github.com/tecnickcom/TCPDF/pull/300#discussion_r532117195

mvorisek: I do not argue the files are large. But `array_indentation` takes at least 100x more time than the other rules. So I think this is an issue.

keradus: as i shared in Fixer repo - if you want, do not hesitate to send a PR to make it faster. if you don't want, don't expect someone will do it for you. same time, you got provided an alternative approach - don't fix autogenerated files that you also do not review.

nicolaasuni (commented):
It looks like that this PR includes breaking changes for older php versions.
I'd rather keep this version as it is and focus on the new tc-lib-pdf library.

mvorisek:
> 
> 
> It looks like that this PR includes breaking changes for older php versions.
> I'd rather keep this version as it is and focus on the new tc-lib-pdf library.

That is correct. TCPDF is hovewer used in a lot of projects and this PR improves maintainability of this project.

I belive dropping support for versions lower than PHP 7.1 is ok for two reasons:
- lower versions of TCPDF can still of course be installed
- PHP 7.1 itself is deprecated and does not receive any updates

mvorisek:
@williamdes what about landing this into WarnockPDF? Maybe even rebase the commits so this is the first commit so the new/fork changes made are easily diffable.

williamdes:
> @williamdes what about landing this into WarnockPDF? Maybe even rebase the commits so this is the first commit so the new/fork changes made are easily diffable.

Good idea, if you can apply https://github.com/wdes/coding-standard that would be awesome !

> Maybe even rebase the commits so this is the first commit so the new/fork changes made are easily diffable.

You can rebase, the history is the same but I did some cleaning recently

mvorisek:
Closing here, if this project is gonna be supported someday, backport the changes from https://github.com/code-lts/WarnockPDF fork

PR #311, opened 2021-01-04
Return cURL output as string for url_exists

matthijsch:
Currently the executed cURL request in the url_exists method outputs the response directly. This is redundant and can make any logging unreadable, slow and large in file size. With this change the response will be returned in curl_exec


matthijsch:
This change seemed to be needed for a specific project. It can be closed.

For your information: In this project the cURL requests needed to be GET and not HEAD, so we removed the line curl_setopt($crs, CURLOPT_NOBODY, true);
Because of this all content was outputted that's why the change of this PR was required.

PR #359, opened 2021-04-03
Remove "U" modifier from regexes

### LINKED ISSUE #267: Empty result for big reports with writeHTML()

mvorisek:
`U` modifier invert greediness which is hard to read, thus remove it

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -16396,7 +16396,7 @@ class TCPDF {
 		// create a special tag to contain the CSS array (used for table content)
 		$csstagarray = '<cssarray>'.htmlentities(json_encode($css)).'</cssarray>';
 		// remove head and style blocks
-		$html = preg_replace('/<head([^\>]*?)>(.*)<\/head>/is', '', $html);
+		$html = preg_replace('/<head([^\>]*?)>(.*?)<\/head>/is', '', $html);
 		$html = preg_replace('/<style([^\>]*?)>([^\<]*?)<\/style>/is', '', $html);
 		// define block tags
 		$blocktags = array('blockquote','br','dd','dl','div','dt','h1','h2','h3','h4','h5','h6','hr','li','ol','p','pre','ul','tcpdf','table','tr','td');
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -16350,15 +16350,15 @@ class TCPDF {
 		$css = array();
 		// get CSS array defined at previous call
 		$matches = array();
-		if (preg_match_all('/<cssarray>([^\<]*)<\/cssarray>/isU', $html, $matches) > 0) {
+		if (preg_match_all('/<cssarray>([^\<]*?)<\/cssarray>/is', $html, $matches) > 0) {
 			if (isset($matches[1][0])) {
 				$css = array_merge($css, json_decode($this->unhtmlentities($matches[1][0]), true));
 			}
-			$html = preg_replace('/<cssarray>(.*?)<\/cssarray>/isU', '', $html);
+			$html = preg_replace('/<cssarray>(.*?)<\/cssarray>/is', '', $html);
 		}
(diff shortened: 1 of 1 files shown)


tcpdf.php
```diff
@@ -16350,15 +16350,15 @@ protected function getHtmlDomArray($html) {
 		$css = array();
 		// get CSS array defined at previous call
 		$matches = array();
-		if (preg_match_all('/<cssarray>([^\<]*)<\/cssarray>/isU', $html, $matches) > 0) {
+		if (preg_match_all('/<cssarray>([^\<]*?)<\/cssarray>/is', $html, $matches) > 0) {
 			if (isset($matches[1][0])) {
 				$css = array_merge($css, json_decode($this->unhtmlentities($matches[1][0]), true));
 			}
-			$html = preg_replace('/<cssarray>(.*?)<\/cssarray>/isU', '', $html);
+			$html = preg_replace('/<cssarray>(.*?)<\/cssarray>/is', '', $html);
```
mvorisek, line 16357: previous behaviour was without the `?`, correct now?

tcpdf.php
```diff
@@ -16396,8 +16396,8 @@ protected function getHtmlDomArray($html) {
 		// create a special tag to contain the CSS array (used for table content)
 		$csstagarray = '<cssarray>'.htmlentities(json_encode($css)).'</cssarray>';
 		// remove head and style blocks
-		$html = preg_replace('/<head([^\>]*)>(.*)?<\/head>/siU', '', $html);
-		$html = preg_replace('/<style([^\>]*)>([^\<]*)<\/style>/isU', '', $html);
+		$html = preg_replace('/<head([^\>]*?)>(.*)<\/head>/is', '', $html);
```
mvorisek, line 16399: I belive this will bring back the #267 issue

@eisberg please test

eisberg: Yes, I tested, with this option, an empty file is generated. This bring back the #267 issue

```
$html = preg_replace('/<head([^\>]*)>(.*)?<\/head>/siU', '', $html); // Works correctly
```

```
$html = preg_replace('/<style([^\>]*?)>([^\<]*?)<\/style>/is', '', $html); // It doesn't work properly, the content is deleted
```

mvorisek, line 16399:
```suggestion
		$html = preg_replace('/<head([^\>]*?)>(.*?)<\/head>/is', '', $html);
```
this works, right? Please also provide a link to full test file.

eisberg: yes, this works

example mini test project:  [test-tcpdf.zip](https://github.com/tecnickcom/TCPDF/files/6271174/test-tcpdf.zip)
Unpack, then run:

```
composer up
php ./test.php
```

mvorisek: resolved in https://github.com/tecnickcom/TCPDF/pull/359#discussion_r608555448

nicolaasuni:
This change seems risky and it is not necessary. I would like to see a test for all cases.

mvorisek:
> 
> 
> This change seems risky and it is not necessary. I would like to see a test for all cases.

`U` modifier is hard to read as it inverts the greediness/lazyness of the quantifiers. This PR removed usage of `U` regex modifier. The function remains completely unchanged. Feel free to test this statement on https://3v4l.org/ website.

nicolaasuni (approved):

PR #362, opened 2021-04-11
Adding missing arial font

Bmorrical:
Can we include arial font in main package so I don't need to include the fork as private repo?


nicolaasuni:
This cannot be included because it is not free software and a Commercial License is required.

PR #364, opened 2021-04-15
Fix Notices on tag styles

sakarikl:


sakarikl:
Ah there already existed a pull request for this

PR #378, opened 2021-05-26
fix warnings for undefined tags for $lineStyle

zymurgic:
Under PHP8, this generated warnings. If a tag is not set, it would throw a warning. Because the downstream usage of the $lineStyle can handle undefined entries, it's safe to leave them unset here.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -18917,13 +18917,30 @@ Putting 1 is equivalent to putting 0 and calling Ln() just after. Default value:
 				$prevlinewidth = $this->GetLineWidth();
 				$this->SetLineWidth($hrHeight);
 
-				$lineStyle = array(
-					'color' => $tag['fgcolor'],
-					'cap'   => $tag['style']['cap'],
-					'join'  => $tag['style']['join'],
-					'dash'  => $tag['style']['dash'],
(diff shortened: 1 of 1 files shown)


williamdes:
I think this would duplicate my work: #350

zymurgic:
> I think this would duplicate my work: #350
I think williamdes may be correct. Both fixes seem to fix the same sort of thing, which is that the tag styles can't be assumed to all be provided at this point.

tcpdf.php
```diff
@@ -18917,13 +18917,30 @@ protected function openHTMLTagHandler($dom, $key, $cell) {
...
-					'phase' => $tag['style']['phase'],
-				);
+				$lineStyle = array();
+                    		if (isset($tag['fgcolor'])) {
+		                        $lineStyle['color'] = $tag['fgcolor'];
+                    		}
+
+                    		if (isset($tag['fgcolor'])) {
+                        		$lineStyle['color'] = $tag['fgcolor'];
+                    		}
```
ChinmayP79, line 18927:
```suggestion
                    		if (isset($tag['fgcolor'])) {
		                        $lineStyle['color'] = $tag['fgcolor'];
                    		}

nicolaasuni (approved):

PR #381, opened 2021-06-04
resolved fopenlocal creating a wrong directory

rashtell:


williamdes (changes requested):
This is not a valid fix

nicolaasuni:
Unfortunately this PR introduces a security bug.
These problems were already fixed in the libraries used by the new tc-lib-pdf.

PR #387, opened 2021-06-11
PHP 8.1: Fix implicit conversion from float to int

mauriciofauth:
Fixes PHP 8.1 deprecation notice: Implicit conversion from float to int loses precision.

- [PHP RFC: Deprecate implicit non-integer-compatible float to int conversions](https://wiki.php.net/rfc/implicit-float-int-deprecate).
- https://github.com/phpmyadmin/phpmyadmin/runs/2806644704#step:10:35

```
PHP Deprecated:  Implicit conversion from float 1.1768666666666663 to int loses precision in /home/runner/work/phpmyadmin/phpmyadmin/vendor/tecnickcom/tcpdf/tcpdf.php on line 4179
PHP Deprecated:  Implicit conversion from float 1.1768666666666663 to int loses precision in /home/runner/work/phpmyadmin/phpmyadmin/vendor/tecnickcom/tcpdf/tcpdf.php on line 4180
```

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -4179,8 +4179,8 @@ class TCPDF {
 			// SHY character will not be printed
 			return (0);
 		}
-		if (isset($this->CurrentFont['cw'][$char])) {
-			$w = $this->CurrentFont['cw'][$char];
+		if (isset($this->CurrentFont['cw'][intval($char)])) {
+			$w = $this->CurrentFont['cw'][intval($char)];
 		} elseif (isset($this->CurrentFont['dw'])) {
 			// default width
 			$w = $this->CurrentFont['dw'];


williamdes (approved):

mvorisek:
I think the fix is done at a wrong place, the char should be always an integer.

mauriciofauth:
> I think the fix is done at a wrong place, the char should be always an integer.

This is done this way other places too. See: https://github.com/tecnickcom/TCPDF/blob/cf8a66c8ec6911bb04f45ccbbf56f2992e6e2139/tcpdf.php#L4685

mvorisek:
Hmm, what `$char` should represent? I still belive it is wrong somewhere else, even the phpdoc at method level say it should be integer.

mauriciofauth:
This stack trace may help find the issue.

```
vendor/tecnickcom/tcpdf/tcpdf.php:4179
vendor/tecnickcom/tcpdf/tcpdf.php:6382
vendor/tecnickcom/tcpdf/tcpdf.php:5956
vendor/tecnickcom/tcpdf/tcpdf.php:3498
vendor/tecnickcom/tcpdf/tcpdf.php:3602
vendor/tecnickcom/tcpdf/tcpdf.php:3269
vendor/tecnickcom/tcpdf/tcpdf.php:3181
```

mauriciofauth:
Anyway, this pull request is a quick fix to remove the deprecation notice, while keeping the current behavior.

nicolaasuni (dismissed):

nicolaasuni:
Please use intval instead of the int casting.

mauriciofauth:
> Please use intval instead of the int casting.

Done!

nicolaasuni (approved):

PR #388, opened 2021-06-15
removed status.txt from font directories, because of filesize

staabm:
looking into the tcpdf vendor folder, the filesize of the vendor/ folder is pretty big.

by far the biggest files are the 2 deleted in this PR.

grafik

I guess these 2 files are pretty unimportant, therefore I suggest this could be deleted (~3 MB of text files)


williamdes:
There is another solution, add them to the  `.gitattributes` file

staabm:
> There is another solution, add them to the `.gitattributes` file

I am fine with that, in case its the preferred solution.

nicolaasuni (approved):

williamdes:
Coyld you rebase please?

PR #389, opened 2021-06-15
fix color being filled to type 0 with php8

sakarikl:
with php8 functionality of following code has changed:  0 != 'text'.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -4025,7 +4025,7 @@ class TCPDF {
 			}
 		}
 		$this->ColorFlag = ($this->FillColor != $this->TextColor);
-		if (($type != 'text') AND ($this->state == 2)) {
+		if (($type != 'text') AND ($this->state == 2) AND $type !== 0) {
 			if (!$ret) {
 				$this->_out($pdfcolor);
 			}


williamdes:
I am not sure to understand why this makes sense for php8 only, can you explain?

sakarikl:
With php 8:  'string' == 0 is false 
With php 7.4 and below 'string' == 0 is true

sakarikl:
https://www.php.net/manual/en/language.operators.comparison.php

More information from there

williamdes:
> With php 8: 'string' == 0 is false
> With php 7.4 and below 'string' == 0 is true

Sure, thank you for the refs, but then changing to `$type !== 'text'` would be a better fix ?

williamdes:
Just confirmed my theory: https://sandbox.onlinephpfunctions.com/

```php
<?php

$type = 0;

if ($type != 'text') {
    echo 'not text 1' . PHP_EOL;
}

if ($type !== 'text') {
    echo 'not text 2' . PHP_EOL;
}
```

is a better fix to use `===`

sakarikl:
yeah, also thought about that. but with that there might be some edge cases that would work differently. with that $type !== 0 only that one specific case is touched.

nicolaasuni (approved):

PR #366, opened 2021-04-20
Increase internal version number to 6.4.1.

paulholden:
Fixes #357.

PR #390, opened 2021-06-23
Update tcpdf.php version info

nagyimre1980:
Update tcpdf.php version info


williamdes:
Good patch !
Goes with #366

nagyimre1980:
> Good patch !
> Goes with #366

#366 only: include/tcpdf_static.php

nicolaasuni (approved):

PR #403, opened 2021-07-21
Leave &NBSP; lowercase when using text-transform

jfdavis:
When using text-transform: uppercase, convert "&NBSP;" back to lowercase "&nbsp;" so that it is recognized as a space by Adobe Acrobat rather than printing out "&NBSP;"

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -17129,6 +17129,7 @@ class TCPDF {
 							}
 						}
 					}
+					$element = preg_replace("/&NBSP;/i", "&nbsp;", $element);
 				}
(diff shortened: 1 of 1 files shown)


jfdavis:
This works for me... if there is a better way to accomplish this in the code, that's fine. :-)

williamdes:
Would you mind sharing an example so some can verify your fix?

jfdavis:
I just noticed that the font definition is not working in 6.4.1 compared to 6.3.2... let me see what's going on with that before submitting samples.

nicolaasuni (approved):

Issue #411, opened 2021-08-03
Text annotations do not appear in the annotations tree


nikitapozdeev:
Hello, thanks for your work.
Correct me if I'm wrong, but it seems to me I found a bug in the behavior of the text annotations.

### Expected Behavior
When adding text annotation to the document, then it should appear in the annotation tree.

### Current Behavior
When adding text annotation, it's not displayed in the annotation tree and this behavior is not controlled by parameters.

This picture illustrates text annotation created in Adobe Acrobat Reader DC
<img width="849" alt="adobe_reader_text_annotation" src="https://user-images.githubusercontent.com/6034438/127991321-4ff3c6ee-85ab-4cf7-b130-31b9d3ba0d1e.png">

This one illustrates text annotation created with TCPDF.
<img width="837" alt="tcpdf_text_annotation" src="https://user-images.githubusercontent.com/6034438/127993480-c5ad68c9-eeff-4377-ae4b-aa60360530ee.png">

This is how I created it
```php
$options = array(
  'Subtype' => 'Text', 
  'Name' => 'Comment', 
  'T' => 'Title example' 
  'Subj' => '', 
  'C' => array(255, 255, 0)
);
$pdf->Annotation(500, 200, 10, 10, 'Hello TCPDF', $options);
```

nicolaasuni:
Thank you for reporting this. TCPDF 6 is now a legacy project and is officially **deprecated**, so no further fixes or features will be merged into this repository. As part of this transition, we are closing older issues. Please migrate to **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** and, if the same problem still occurs there, open a new issue with updated reproduction steps. For more context, see the pinned announcement: [Thank You - and the Future of TCPDF](https://github.com/tecnickcom/TCPDF/issues/867).

PR #412, opened 2021-08-03
fix text-annotation state options

nikitapozdeev:
Hello.
Fix for #411 

I did not change the current conditions, just turned off the setting of the Statе and StateModel parameters for text annotations, if they are not explicitly specified.
According to PDF specification 1.7, these parameters are optional.


williamdes (approved):
Looks good, from what I understand if it is not specified the code must not put default values

nicolaasuni (approved):

Issue #419, opened 2021-08-16
sync up functions names for case


gemal:
so the function is called `setHeaderData`
https://github.com/tecnickcom/TCPDF/blob/main/tcpdf.php#L3325

but the examples all seems to use: `SetHeaderData`
https://github.com/tecnickcom/TCPDF/blob/main/examples/example_038.php#L41

would be nice if all of the "setXXX" functions where synced up

gemal:
php functions are case insensitive but just for consistency in the project

williamdes:
Good idea, could you make a PR ?

gemal:
what is the correct use?
`setHeaderData` or `SetHeaderData`

I personally vote for `setHeaderData`

williamdes:
`setHeaderData` because it is what the real function name is

gemal:
I've started with the `set` functions
Later on we can look at others like `GetNumChars` functions

gemal:
Let me know on how to proceed. Are they any automated tests?

williamdes:
> Let me know on how to proceed. Are they any automated tests?

There is no tests on this project but all the examples are run on all possible php versions

I bet a tool like phpstan found find out all the non well cased functions isuses.

Best is to create a PR with a commit fixing all the casing of the functions in examples

PR #420, opened 2021-08-16
fix Set to set in functions. Fixes up #419 

gemal:


tcpdf.php
```diff
@@ -2604,9 +2604,9 @@ public function SetMargins($left, $top, $right=-1, $keepmargins=false) {
 	 * @param float $margin The margin.
 	 * @public
 	 * @since 1.4
-	 * @see SetTopMargin(), SetRightMargin(), SetAutoPageBreak(), SetMargins()
+	 * @See setTopMargin(), setRightMargin(), setAutoPageBreak(), setMargins()
 	 */
-	public function SetLeftMargin($margin) {
+	public function setLeftMargin($margin) {
```
williamdes, line 2609: hum, this is not right, you may not change existing function definitions

gemal: why not? PHP function calling is case insensitive

williamdes: My point is that the people using the project will not like this change, maybe it will break something
I would recommend you to revert thoes changes because you will risk that the owner of the project does not merge your changes :/

gemal: I don't see how this could break anything, but I created a new pull request

PR #421, opened 2021-08-16
change Set to set. Fixes #419

### LINKED ISSUE #419: sync up functions names for case

gemal:


williamdes (approved):
LGTM

williamdes:
@gemal the functions are still starting with a `S` and not a `s`, was this sync a good thing ?

gemal:
IMHO they should be changed from S to s

PR #434, opened 2021-09-16
Fixes check whether curl can be used

pjdevries:
Fixes check whether curl can be used in method TCPDF_STATIC::fileGetContents().


pjdevries:
> 
> 
> [CLA assistant check](https://cla-assistant.io/tecnickcom/TCPDF?pullRequest=434)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/tecnickcom/TCPDF?pullRequest=434) before we can accept your contribution.
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/tecnickcom/TCPDF?pullRequest=434) it.

I'm very sorry, but I'm not going to sign anything. You can take or leave my contribution, whichever suits you best

include/tcpdf_static.php
```diff
@@ -1981,7 +1981,7 @@ public static function fileGetContents($file) {
 			    return $ret;
 			}
 			// try to use CURL for URLs
-			if (!ini_get('allow_url_fopen')
+			if (ini_get('allow_url_fopen')
```
williamdes, line 1984: For me this has nothing to do with checking curl is available or not

pjdevries: I guess you're right. The issue should probably be titled 'Fixes if url's can be accessed by curl'. Be me guest and change it. The fix is valid nevertheless though.

williamdes: I can not understand why `allow_url_fopen` is checked since it has nothing to do with curl ?
Or is it to make a priority on `url_fopen` over curl ?

pjdevries: To be honest, I don't know. The only thing I know is that it works as expected, after making this change change :)

pjdevries: It's not my project and the code is far to complex (understatement) to analyze how this issue could be fixed in a way you see fit. Besides, it does not seem worth it for a project that is no longer maintained, according to the README. So I can do nothing more than repeat myself by saying: 'It works as expected, after making this change change'. So I'll keep using my fork :)

williamdes: Yup understood, can you close this PR then ?

Issue #371, opened 2021-04-28
TCPDF inside a webPhar does not work (broken since 6.2.19)


jamjamg:
I' m using TCPDF inside a web application packaged as a PHAR file. Until now I was stuck to version 6.2.17 because for me this was the last working version.  Since 6.2.19 font files cannot be included anymore (tested all versions from 6.2.17 to 6..4.1):
```
Uncaught PHP Exception Exception: "PDF generation failed: Could not include font definition file: helvetica" at phar:///../website.phar/src/.../MyPdfImplementation.php line 123 {"exception":"[object] (Exception(code: 0): PDF generation failed: Could not include font definition file: helvetica at phar:///../website.phar/src/.../MyPdfImplementation.php:123)"} []
```

For PHP 8 support I decided to update to 6.4.1 and digged a little deeper. I got it working again by allowing `phar://` and `file://` urls in `tcpdf_static.php`:

```
public static function file_exists($filename) {
    if (preg_match('|^https?://|', $filename) == 1) {
        return self::url_exists($filename);
    }
    if (stripos($filename, 'phar://') !== false || stripos($filename, 'file://') !== false) {
        // allow phar:// and file:// urls, so TCPDF itself can be inside a 'webPhar'
        return @file_exists($filename);
    }
    if (strpos($filename, '://')) {
        return false; // only support http and https wrappers for security reasons
    }
    return @file_exists($filename);
}
```
The breaking change was https://github.com/tecnickcom/TCPDF/commit/9f0350dca7ce43c264c00bdce5fa493b9e358c94 .

And I'm not sure if my patch would break security fix https://github.com/tecnickcom/TCPDF/commit/1861e33fe05f653b67d070f7c106463e7a5c26ed .

pcastelovigo:
thanks @jamjamg

nicolaasuni:
Thank you for reporting this. TCPDF 6 is now a legacy project and is officially **deprecated**, so no further fixes or features will be merged into this repository. As part of this transition, we are closing older issues. Please migrate to **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** and, if the same problem still occurs there, open a new issue with updated reproduction steps. For more context, see the pinned announcement: [Thank You - and the Future of TCPDF](https://github.com/tecnickcom/TCPDF/issues/867).

PR #438, opened 2021-09-22
Allow phar urls usage in file_exists method to fix font loading from phar app

petitj:
As it is referred by issue  #371, TCPF breaks inside phar because the internal file_exist method disallow anything except local, http and https protocols. Whenever TCPDF tries to load a font, it then fails with message "Could not include font definition file: helvetica". This P.R solves the problem by including the phar protocol to the list (phar://).

### DIFF
--- a/include/tcpdf_static.php
+++ b/include/tcpdf_static.php
@@ -1898,8 +1898,8 @@ class TCPDF_STATIC {
 		if (preg_match('|^https?://|', $filename) == 1) {
 			return self::url_exists($filename);
 		}
-		if (strpos($filename, '://')) {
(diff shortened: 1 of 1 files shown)


nicolaasuni:
Unfortunately this change appears to introduce a security issue related to the phar file format.
A better implementation should be introduced in [tc-lib-file](https://github.com/tecnickcom/tc-lib-file) and https://github.com/tecnickcom/tc-lib-pdf.

petitj:
Hi and thanks for the review. 
I'm curious to know what security issue is introduced by the use of phar protocol as i can not see on the top of my head how it can be worse than http. Do you have any details or pointers ?

PR #462, opened 2021-12-30
lowercase s in function names

gemal:
more fix for #419


nicolaasuni (approved):

williamdes:
For me this is a BC break, right?

gemal:
Php functions are case insensitive. So setX and Setx is the same

PR #466, opened 2022-01-07
GetCssBorderStyle Has Problem When !important Is Specified

oleibman:
This is, in particular, a problem for PhpSpreadsheet. See https://github.com/PHPOffice/PhpSpreadsheet/issues/1164. TCPDF thinks that there can be 0, 1, 2, or 3 components to shorthand border style, but there can be more, e.g. !important. Logic is rearranged to allow for more than 3. A new example file is added to demonstrate that the fix works correctly.

### DIFF
--- a/examples/example_067.php
+++ b/examples/example_067.php
@@ -1,0 +1,151 @@
+<?php
+//============================================================+
+// File name   : example_067.php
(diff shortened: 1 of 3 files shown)


oleibman:
Problems with userid. Will retry.

PR #473, opened 2022-01-25
Fix Infinite Loop in Multicell with Auto Page Breaks Off

ssigwart:
The updated places that used `AcceptPageBreak` assumed that a page break was added and increased X by the margin.  However, if the break wasn't added, it would put the text further to the right to the point that the width because so small or negative and no characters fit, causing an infinite loop.

# Example File
```php
<?php

require_once('tcpdf_include.php');

class MYPDF extends TCPDF
{
	public function crash()
	{
		$width = 140;
		$val = '';
		for ($i = 1; $i <= 100; $i++)
			$val .= "Line $i\n";
		$this->MultiCell($width, 10, $val, 1, 'L');
	}
}

$pdf = new MYPDF('P', 'mm', 'A4', true, 'UTF-8', false);
$pdf->setMargins(10, 10, 10);
$pdf->setCellMargins(10, 0, 10, 0);
$pdf->setAutoPageBreak(false);
$pdf->setFont('helvetica', '', 12);
$pdf->AddPage();
$pdf->crash();
$pdf->Output('file.pdf', 'I');

```

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -6495,14 +6495,16 @@ class TCPDF {
 				$shy = false;
 				// account for margin changes
 				if ((($this->y + $this->lasth) > $this->PageBreakTrigger) AND ($this->inPageBody())) {
-					$this->AcceptPageBreak();
-					if ($this->rtl) {
-						$this->x -= $margin['R'];
-					} else {
-						$this->x += $margin['L'];
+					if ($this->AcceptPageBreak())
+					{
+						if ($this->rtl) {
+							$this->x -= $margin['R'];
+						} else {
+							$this->x += $margin['L'];
+						}
+						$this->lMargin += $margin['L'];
+						$this->rMargin += $margin['R'];
 					}
-					$this->lMargin += $margin['L'];
-					$this->rMargin += $margin['R'];
 				}
 				$w = $this->getRemainingWidth();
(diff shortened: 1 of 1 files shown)


tcpdf.php
```diff
@@ -6495,14 +6495,16 @@ public function Write($h, $txt, $link='', $fill=false, $align='', $ln=false, $st
...
 				// account for margin changes
 				if ((($this->y + $this->lasth) > $this->PageBreakTrigger) AND ($this->inPageBody())) {
-					$this->AcceptPageBreak();
-					if ($this->rtl) {
-						$this->x -= $margin['R'];
-					} else {
-						$this->x += $margin['L'];
+					if ($this->AcceptPageBreak())
+					{
+						if ($this->rtl) {
```
williamdes, line 6500: should this be reached if AcceptPageBreak is false ?

ssigwart: I don't think this code should be run if `AcceptPageBreak` is false.  My understanding is that `AcceptPageBreak` will return false if a page break is not added.  If a page break is not added, there should be no need to adjust the margin.

nicolaasuni (approved):

ssigwart:
Thanks for merging this, @nicolaasuni!

PR #474, opened 2022-01-25
Add page cache option to save memory

ssigwart:
Running example_067.php with and without the cache enabled uses about 26MB without vs 17MB with the cache.

This adds the `usePageCacheFile` option which allow you to move page data to a temporary file when not in use.


examples/index.php
```diff
@@ -85,6 +85,7 @@
 <li>Text stretching and spacing (tracking/kerning): [<a href="example_063.php" title="PDF [new window]" target="_blank">PDF</a>]</li>
 <li>No-write page regions: [<a href="example_064.php" title="PDF [new window]" target="_blank">PDF</a>]</li>
 <li>PDF/A-1b (ISO 19005-1:2005) document: [<a href="example_065.php" title="PDF [new window]" target="_blank">PDF</a>]</li>
+<li>Using page cache to reduce memory usage: [<a href="example_066.php" title="PDF [new window]" target="_blank">PDF</a>]</li>
```
williamdes, line 88:
```suggestion
<li>Using page cache to reduce memory usage: [<a href="example_067.php" title="PDF [new window]" target="_blank">PDF</a>]</li>
```

tcpdf.php
```diff
@@ -174,6 +174,24 @@ class TCPDF {
 	 */
 	protected $pages = array();
 
+	/**
+	 * Page cache file
+	 * @protected
```
williamdes, line 179:
```suggestion
	 * @protected
	 * @var ressource|null
```

tcpdf.php
```diff
@@ -174,6 +174,24 @@ class TCPDF {
...
+	 */
+	protected $pageCacheFile = null;
+
+	/**
+	 * Page cache file is readonly
+	 * @protected
```
williamdes, line 185:
```suggestion
	 * @protected
	 * @var bool
```

tcpdf.php
```diff
@@ -7866,6 +7889,49 @@ public function _destroy($destroyall=false, $preserve_objcopy=false) {
 		}
 	}
 
+	/** Page cache reference counts */
+	protected static $pageCacheRefCnts = [];
+
+	/**
+	 * Update page cache file
+	 *
+	 * @param int $memSizeInBytes Memory size to use in bytes
```
williamdes, line 7898:
```suggestion
	 * @param int $memSizeInBytes Memory size to use in bytes
	 * @return void
```

tcpdf.php
```diff
@@ -7866,6 +7889,49 @@ public function _destroy($destroyall=false, $preserve_objcopy=false) {
...
+	 */
+	public function usePageCacheFile($memSizeInBytes)
+	{
+		$this->pageCacheFile = fopen('php://temp/maxmemory:' . $memSizeInBytes, 'w+');
+		if ($this->pageCacheFile === false)
+			$this->pageCacheFile = null;
+		else
+			self::$pageCacheRefCnts[(int)$this->pageCacheFile] = 1;
```
williamdes, line 7906:
```suggestion
		if ($this->pageCacheFile === false) {
			$this->pageCacheFile = null;
		} else {
			self::$pageCacheRefCnts[(int)$this->pageCacheFile] = 1;
		}
```

tcpdf.php
```diff
@@ -7866,6 +7889,49 @@ public function _destroy($destroyall=false, $preserve_objcopy=false) {
...
+
+	/**
+	 * Close page cache file
+	 */
+	private function _closePageCacheFile()
+	{
+		if ($this->pageCacheFile !== null)
+		{
+			self::$pageCacheRefCnts[(int)$this->pageCacheFile]--;
+			if (self::$pageCacheRefCnts[(int)$this->pageCacheFile] == 0)
```
williamdes, line 7917: please use `{` for your if blocks

ssigwart:
Thanks for reviewing it, @williamdes!  I think I got all the changes in and updated all the `if`s.

tcpdf.php
```diff
@@ -7866,6 +7891,52 @@ public function _destroy($destroyall=false, $preserve_objcopy=false) {
...
+	/**
+	 * Close page cache file
+	 */
+	private function _closePageCacheFile()
+	{
+		if ($this->pageCacheFile !== null)
+		{
+			self::$pageCacheRefCnts[(int)$this->pageCacheFile]--;
+			if (self::$pageCacheRefCnts[(int)$this->pageCacheFile] == 0) {
+				@fclose($this->pageCacheFile); // Suppress error since this might be the end of the PHP script
```
williamdes, line 7922: A is ressource check might remove the need of silencing an error

tcpdf.php
```diff
@@ -7866,6 +7891,52 @@ public function _destroy($destroyall=false, $preserve_objcopy=false) {
...
+			}
+			$this->pageCacheFile = null;
+			$this->pageCacheIndex = [];
+		}
+	}
+
+	/** Handle cloning */
```
williamdes, line 7929: Why is this needed, it feels a bit incomplete or non explained

tcpdf.php
```diff
@@ -20918,8 +20996,45 @@ protected function setPageBuffer($page, $data, $append=false) {
...
+						$origPageCacheFile = $this->pageCacheFile;
+						self::$pageCacheRefCnts[(int)$this->pageCacheFile]--;
+						$this->pageCacheFile = fopen('php://temp', 'w+');
+						if ($this->pageCacheFile === false) {
+							$this->pageCacheFile = null;
+						} else {
+							self::$pageCacheRefCnts[(int)$this->pageCacheFile] = 1;
+							// Copy data
+							fseek($origPageCacheFile, 0, SEEK_SET);
+							while (($copyContent = fread($origPageCacheFile, 8192)) != false)
```
williamdes, line 21017: This is missing {

Could you add a bit more comments on what each step does?

tcpdf.php
ssigwart: Good suggestion.  I ran a test and that seems like a better way to do this.  I updated it.

tcpdf.php
ssigwart: I updated the comment.  It's so that if you clone the object, you don't have two objects writing (and overwriting each other) to the same file.

tcpdf.php
ssigwart: I added the brace and added comments.

tcpdf.php
```diff
@@ -7866,6 +7891,60 @@ public function _destroy($destroyall=false, $preserve_objcopy=false) {
 		}
 	}
 
+	/** Page cache reference counts */
+	protected static $pageCacheRefCnts = [];
```
williamdes, line 7895:
```suggestion
	protected static $pageCacheRefCnts = array();
```
Old php compatibility

tcpdf.php
```diff
@@ -21201,7 +21332,9 @@ public function deletePage($page) {
 			return false;
 		}
 		// delete current page
-		unset($this->pages[$page]);
+		if (isset($this->pages[$page])) {
+			unset($this->pages[$page]);
```
williamdes, line 21336: A bit too much indentation
Same below

tcpdf.php
```diff
@@ -174,6 +174,26 @@ class TCPDF {
...
+	/**
+	 * Page cache file is readonly
+	 * @protected
+	 * @var bool
+	 */
+	protected $roPageCacheFile = false;
+
+	/**
+	 * Page cache index
+	 * @protected
```
williamdes, line 193:
```suggestion
	 * @protected
	 * @var array<int,int>
```

williamdes (dismissed):
This code is crystal clear, thank you!

tcpdf.php
ssigwart: I updated this.

tcpdf.php
ssigwart: Sorry, I don't understand this comment.  It seems to match the other if statements.

williamdes (approved):
💯

ssigwart:
@williamdes, I assume someone else will merge this at some point, right?  I don't seem to have permission.

williamdes:
> @williamdes, I assume someone else will merge this at some point, right? I don't seem to have permission.

Yes, when Nicolas will have some time he will probably review this :)
Now we need to be patient

ssigwart:
Sounds good.  Thank you.

tcpdf.php
```diff
@@ -7866,6 +7892,60 @@ public function _destroy($destroyall=false, $preserve_objcopy=false) {
 		}
 	}
 
+	/** Page cache reference counts */
+	protected static $pageCacheRefCnts = array();
```
williamdes, line 7896: can this be non static ?

ssigwart: I intentionally have it static so that it can be shared when you clone it. The PR implements copy-on-write, so if you clone the PDF, it will share the cache until an edit is made.

I've been including this patch on production since 2019 and I haven't noticed any issues with it, so I feel like it's safer to leave this as-is, but if you want, I can remove the shared cache pages and make this non-static.

williamdes: I think it could cause issues to users that do not expect things to be cached between pdfs

ssigwart: Do you mean people using the library? They would have to enable the page cache with `usePageCacheFile(...)` and even if they do, I have the code internally handling the caching between PDFs, so I don't think it would cause issues or be something users even know about.

If you are referring to people developing the library, that's where I think it might cause confusion, but I think I can handle that by writing more detailed comments.

If the `static` part is something you want to change, but still keep caching across PDFs, I could implement that with a new object for reference counting and since objects are shared by reference, it would act very similarly to the static variable.

williamdes: I think that if you can implement a version with objects that would be better than a static variable. @MauricioFauth maybe you would have input on this feature and the use of statics?

ssigwart: I added e5a3a7ce85c5c18066fcff7f87c3cf3f25fca2ad, which does it with an object instead. I'm not as confident in this change since it hasn't had years of real world testing, but it seems correct.  Personally, I think this might be a little more confusing for library developers. With the static variable, I think it's more obvious that it's used to share data than this object by reference approach.

williamdes: Thanks, I hope someone with merge power can review this some time soon
The commit can be reverted if the static approach is preferred by the reviewer

ssigwart:
I rebased this to fix the merge conflicts with the examples.

ssigwart:
I also updated this so it passes the unit tests, including on PHP 5.3.

williamdes:
> I also updated this so it passes the unit tests, including on PHP 5.3.

Thank you !
We will have to wait on @nicolaasuni to a review

nicolaasuni:
While this is a nice feature to have, I am not keen to accept new features as the tc-lib-pdf is already working and missing only (even if very important) HTML/CSS/SVG/Javascript support.

ssigwart:
Thanks, @nicolaasuni. That makes sense. I'm looking forward to the production release of tc-lib-pdf to see how it works with some of the large CSVs I generate.

PR #509, opened 2022-04-13
Fix garbage collection

deguif:
When registering a shutdown function referencing to the tcpdf instance, it prevents this instance to garbage collected.
The desctructor call `destroy(true)`.


williamdes (dismissed):
How did you find that ?

deguif:
By investigating a long running script that generates hundred of PDFs. When debugging inside the `__destruct` method, I was surprised to see it was never called, until the script ended (even when `gc_collect_cycles` was called after each pdf generation).

williamdes:
> By investigating a long running script that generates hundred of PDFs (with memory increasing). When debugging inside the `__destruct` method, I was surprised to see it was never called, until the script ended (even when `gc_collect_cycles` was called after each pdf generation).

Merci !
Could this be proved by some code written in a file like https://github.com/tecnickcom/TCPDF/blob/main/examples/example_066.php ?

nicolaasuni:
The _destroy function also deletes temporary files.
Removing this will leave zombie files?

deguif:
@nicolaasuni, that should not be the case, as the `_destroy` method is called on the destructor. That way it will be destroyed natively by the PHP engine, when object instance got out of scope (not referenced anymore).

deguif:
Just rebased the branch

williamdes (approved):

deguif:
Any way to get this one merged?

williamdes:
@nicolaasuni does reviews from time to time, months apart and randomly. So it's up to us to wait once a happy day arrives :)

nicolaasuni (approved):

Issue #303, opened 2020-11-26
Images in HTML rtl not aligned correctly


julienfelixweb:
Hi I'm meeting this issue, is there a simple fix to solve this ?

![RTL](https://user-images.githubusercontent.com/770119/100354342-50409b80-2ff0-11eb-962e-579dde09be55.png)

Thanks

OTW-DEV-1:
I am also facing same problem.
![rtl images](https://user-images.githubusercontent.com/73678532/101555822-384b1d80-39db-11eb-839f-873f82fb4804.png)

alibarrak:
you need to split your HTML (setRTL(false) before images)

for example image is in the middle:
`
$pdf->setRTL(true);
$html = 'some rtl text......';
$pdf->writeHTML($html, true, false, true, false, '');

$pdf->setRTL(false);
$html = '<img ....>';
$pdf->writeHTML($html, true, false, true, false, '');

$pdf->setRTL(true);
$html = 'some rtl text.....';
$pdf->writeHTML($html, true, false, true, false, '');
`

julienfelixweb:
Thanks @alibarrak for this work around, unfortunately it won't fix it in my case.
But it demonstrates that the img tag is not well handle in RTL.
I'm sure it's an easy fix but I can't figure out where it is. It seams that in RTL images are systematically push to the right and the amount of this translation is equal to the width.

julienfelixweb:
I did found a fix which I'm still testing. see #165

ctadri:
I have found the exact same issue after the change from https://github.com/tecnickcom/TCPDF/issues/165, in my use case I am able to fix it with this code

![image](https://user-images.githubusercontent.com/62693299/128173398-9d7b91e8-daed-4265-b357-9ca719a59d14.png)

This is just a quickfix, which could solve https://github.com/tecnickcom/TCPDF/issues/303 but maybe creates more problems with https://github.com/tecnickcom/TCPDF/issues/165

williamdes:
Seems to be fixed with #510

nicolaasuni:
[...]

PR #510, opened 2022-04-13
Fix image abscissa when in RTL

atomiix:
As reported in #303, if you try [example_018.php](https://github.com/tecnickcom/TCPDF/blob/main/examples/example_018.php), you'll see:
Capture d’écran 2022-04-13 à 16 03 03

This fix comes from [@ctadri](https://github.com/ctadri), but then [example_051.php](https://github.com/tecnickcom/TCPDF/blob/main/examples/example_051.php) doesn't work anymore when in RTL (by modifying [lang/eng.php](https://github.com/tecnickcom/TCPDF/blob/main/examples/lang/eng.php#L36)). However, by setting the `palign` parameter of the function `Image()` to `'C'`, you can make it work back.

More generally, I think it's easier to tweak a direct call to `Image()` to place the image where you want according to the language direction than to tweak the html to have the image aligned as you want.


tcpdf.php
```diff
@@ -7230,7 +7230,11 @@ public function Image($file, $x=null, $y=null, $w=0, $h=0, $type='', $link='', $
 		} elseif ($palign == 'R') {
 			$ximg = $this->w - $this->rMargin - $w;
 		} else {
-			$ximg = $x;
+			if ($this->rtl) {
+				$ximg = $x - $w;
+			} else {
+				$ximg = $x;
+			}
```
williamdes, line 7237:
```suggestion
			$ximg = $this->rtl ? $x - $w : $x;
```

williamdes (dismissed):

williamdes (approved):

williamdes (approved):

nicolaasuni (approved):

PR #516, opened 2022-05-03
Php8.0 fixes

pdfgeneratorapi:
checking if $chrw has at least one item


williamdes:
Why did you close your PR ?
Can you provide an example that creates this error ?

bruno-farias:
Will include more fixes in a next PR

Issue #470, opened 2022-01-16
Warnings if strings containe special chars like äöüß in german language


kruegge82:
we using php 7.4 and TCPDF via composer latest version 6.4.4

[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  array_map(): Expected parameter 2 to be an array, bool given in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 2004
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  array_merge(): Expected parameter 2 to be an array, null given in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 2012
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  Invalid argument supplied for foreach() in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 1844
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  max(): Array must contain at least one element in vendor/tecnickcom/tcpdf/tcpdf.php on line 6398
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  array_map(): Expected parameter 2 to be an array, null given in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 1783
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  count(): Parameter must be an array or an object that implements Countable in vendor/tecnickcom/tcpdf/tcpdf.php on line 6402
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  array_merge(): Expected parameter 1 to be an array, null given in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 2012
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  array_merge(): Expected parameter 1 to be an array, null given in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 2012
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  array_merge(): Expected parameter 1 to be an array, null given in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 2012
[16-Jan-2022 10:58:59 Europe/Berlin] PHP Warning:  array_map(): Expected parameter 2 to be an array, bool given in vendor/tecnickcom/tcpdf/include/tcpdf_fonts.php on line 2004
[...]

PR #517, opened 2022-05-04
Bug fix:  returning false

KevinVG:


include/tcpdf_fonts.php
```diff
@@ -2001,7 +2001,7 @@ public static function UTF8StringToArray($str, $isunicode, &$currentfont) {
 		if ($isunicode) {
 			// requires PCRE unicode support turned on
 			$chars = TCPDF_STATIC::pregSplit('//','u', $str, -1, PREG_SPLIT_NO_EMPTY);
-			$carr = array_map(array('TCPDF_FONTS', 'uniord'), $chars);
+			$carr = array_map(array('TCPDF_FONTS', 'uniord'), is_array($chars) ? $chars : array());
```
williamdes, line 2004: Can you provide an example code to trigger this?

julienbdx (approved):

include/tcpdf_fonts.php
sgiehl: I think the problem is related to https://github.com/tecnickcom/TCPDF/issues/470
When providing a non utf8 string the `preg_split` might return `false`. causing the array_map to throw a warning or error. But I'm not sure if this is the optimal solution, as it would simply remove the text from the pdf...

nicolaasuni (approved):

nicolaasuni (approved):

nicolaasuni:
@KevinVG please sign the license/cla to be able to merge this.

nicolaasuni:
Closing this as CLS was not signed and the issue was solved by:
https://github.com/tecnickcom/TCPDF/commit/e3d45a685352419e02a757f729f3ee153ccf217c

PR #518, opened 2022-05-07
Output error

BondorD:
Output(,'F') error and how I managed to fix it

### DIFF
--- a/include/tcpdf_static.php
+++ b/include/tcpdf_static.php
@@ -1834,11 +1834,11 @@ class TCPDF_STATIC {
 	 * @public static
 	 */
 	public static function fopenLocal($filename, $mode) {
-		if (strpos($filename, '://') === false) {
+		/* if (strpos($filename, '://') === false) {
 			$filename = 'file://'.$filename;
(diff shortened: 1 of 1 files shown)


sagehenstudio:
Curious, why throw baby out with the bathwater? This method will throw errors in certain instances. What was your error? Have you confirmed the file in question is actually in the location and accessible (e.g. not blocked by something like PrivateTmp)? Could you provide more background as to why you suggest these changes?

DuckDensity:
I ended up making the same edit during development.

When running on the command line (in Windows environment) and executing a php script stored on a network device using a UNC path, this code prevents the generated PDF file from being saved to the same folder system.
A work around is to use a mapped drive letter, rather than UNC Path.

This error is:
`Warning: fopen(file://\\Server\Share\Path\File ....) Failed to open stream`

Not too sure of the security implications here, but when working on CLI with a known file system it should be safe to save?
Perhaps this function needs to behave differently for CLI compared to web server environment?

nicolaasuni:
The removed section is there for security reasons, so I can't merge this.

PR #530, opened 2022-06-21
Support Apache 2.4 directives in .htaccess file

SonOfHarris:
Update the .htaccess files to support Apache 2.4 directives with backwards compatibility for Apache 2.2.

See: https://httpd.apache.org/docs/2.4/upgrading.html#run-time


williamdes (approved):

shane-smith:
I would also like to see this change merged.

nicolaasuni (approved):

PR #489, opened 2022-02-25
[FIX] PDF417 produces corrupt output #488

bayramarslan:
[FIX] PDF417 produces corrupt output [https://github.com/tecnickcom/TCPDF/issues/488](url)

PR #534, opened 2022-06-30
FIX: PDF417 corrupt output problem, solution set bcmath scale parameter to zero

bayramarslan:
FIXES #489


williamdes (approved):

nicolaasuni:
The barcode methods in this library are obsolete and unmaintained.
Please use instead tc-lib-barcode.

nicolaasuni:
The scale parameter was added in PHP 7.2 and this change will break backward compatibility.
TCPDF is compatible with PHP 5.3.0

nicolaasuni (approved):

nicolaasuni:
@bayramarslan please sign the license/cla to merge this.

Issue #537, opened 2022-07-05
Project maintenance


theofidry:
Hi @nicolaasuni,

I was wondering if you would consider adding new maintainers to the project. I think a few people are looking to help, e.g. @mvorisek and @williamdes were a bit active last year and I think in general there is quite a few attempts at contributing. I would also be happy to help: I have little personal time for it but since my company is using it I can definitely allocate some time for at the very least helping with the reviews & merging.

I however understand it's not an easy decision, and there is always the fear of the new maintainers going in a very different direction, messing things up, being too aggressive with refactorings or migrating etc.
If this is the case, I would highly recommend the approach adopted by @\fzaninotto for https://github.com/fzaninotto/Faker or myself for https://github.com/humbug/humbug which is archiving the project and pointing to an alternative, which could also be a fork.

This package can boast of a 35k daily download and is definitely one if not the most popular PDF generator in the PHP community. I do use it for a couple of professional projects as well and personally I would love to see it be active again one way or another and propose changes to come back more in line with modern PHP whilst providing an acceptable migration path and BC compatibility as much as possible given its wide usage.

williamdes:
> would consider adding new maintainers to the project. I think a few people are looking to help, e.g. @mvorisek and @williamdes were a bit active last year and I think in general there is quite a few attempts at contributing. 

I am willing to continue my daily issue review and PR reviews
I would definitely need some communication from the project owner
I had started a fork as some users have done, just to merge bug fixes and have them in production

This project is so huge and used everywhere it is critical it stays maintained as long as possible
But without communication it will take another path some day and I really do not know what to expect
We use it at work, some laravel libs use it
And at @phpmyadmin we use it
I even do the Debian/Ubuntu packaging for tcpdf: https://tracker.debian.org/pkg/tcpdf

So please @nicolaasuni give us some hope 🙏🏻 
We can invest time to keep this afloat

williamdes:
Ref https://github.com/code-lts/WarnockPDF/discussions/1#discussioncomment-537344

managerrojo:
hi guys im working in a project that use it this Library and in the error Log show me:

Array and string offset access syntax with curly braces is no longer supported in /var/www/html/vendor/rafikhaceb/tcpdi/tcpdi_parser.php on line 818

818 : $frag = $data{$offset} . @$data{$offset+1} . @$data{$offset+2} . @$data{$offset+3};

since this way is deprecate after 7.4 a this new way : 

[...]

PR #541, opened 2022-07-26
Allow for bar width reduction

ro0NL:
Hi,

See https://barcode-test.com/barcode-advice/bar-width-reduction/ for more information

See also https://sourceforge.net/p/tcpdf/discussion/435311/thread/e00f50f8/#756c, which ultimately led to this fix

I can confirm _BWR_ on our side is also needed to account for "press gain compensation"

We are currenly applying this patch for 2-3 years in prod, and if it could be merged upstream, well, i'd be super happy :)

Thoughts?


williamdes:
Could you also add an example file in `examples` like https://github.com/tecnickcom/TCPDF/blob/main/examples/example_066.php ?

ro0NL:
@williamdes i've updated existing example_027, which seemed best for general discoverability

`.5` as an example value should give a good impression what this option does:

image

williamdes (approved):
Thank you, this is really awesome!

ro0NL:
cc @nicolaasuni (yes being a project maintainer means random ppl will ping you)

ro0NL:
closing while #537 remains open

PR #548, opened 2022-08-31
fix: php 8.1 notices

jmaumene:
PHP Deprecated:  Implicit conversion from float 31.5 to int loses precision in /var/www/html/vendor/tecnickcom/tcpdf/include/barcodes/qrcode.php on line 891
PHP Deprecated:  Implicit conversion from float 3.5 to int loses precision in /var/www/html/vendor/tecnickcom/tcpdf/include/barcodes/qrcode.php on line 895


williamdes (approved):

jtvanschagen (approved):

jmaumene:
there is a bug with the CLA, I signed it the first day and started over several times.

andreasvonburg:
Would be great if this could be merged.

marinaglancy:
hi, this is a correct patch and solves the problem for us (Moodle) as well. Can it please be merged?

williamdes:
Hi @nicolaasuni 
Would you mind having a look ?

Issue #561, opened 2022-11-14
Multi-byte characters in filename are silently discarded during output


paulholden:
When a filename contains multi-byte characters (as in the description of https://tracker.moodle.org/browse/MDL-76273) they are silently discarded by the `Output()` method

It would be good to encode them, such that they are preserved

Similar to issue #514

PR #562, opened 2022-11-14
Multi-byte character support for filename during output (#561).

paulholden:
Resolution for https://github.com/tecnickcom/TCPDF/issues/561

When a filename contains multi-byte characters (as in the description of https://tracker.moodle.org/browse/MDL-76273) they are silently discarded by the Output() method

It would be good to encode them, such that they are preserved


tcpdf.php
```diff
@@ -7711,7 +7708,8 @@ public function Output($name='doc.pdf', $dest='I') {
 					header('Pragma: public');
 					header('Expires: Sat, 26 Jul 1997 05:00:00 GMT'); // Date in the past
 					header('Last-Modified: '.gmdate('D, d M Y H:i:s').' GMT');
-					header('Content-Disposition: inline; filename="'.basename($name).'"');
+					header('Content-Disposition: inline; filename="' . rawurlencode(basename($name)) . '"; ' .
+						'filename*=U-TF8\'\'' . rawurlencode(basename($name)));
```
Fighter456, line 7712:
```suggestion
						'filename*=UTF-8\'\'' . rawurlencode(basename($name)));
```

Fighter456 (changes requested):
Otherwise LGTM

tcpdf.php
paulholden: Wow, can't believe I missed that - thanks @Fighter456 for your quick response :+1:

Fighter456 (approved):

nicolaasuni (approved):

PR #572, opened 2022-12-10
Extract method to get TCPDF producer

MauricioFauth:
Wraps `TCPDF_STATIC::getTCPDFProducer()` calls with a private method and adds an empty private property that can be used to change the producer line. The private property can be changed with the reflection API.

Both method and property are declared as private to avoid BC breaks.

The idea is to help with testing PDF generation, by using a predictable producer line.

- Related to https://github.com/tecnickcom/TCPDF/pull/85


williamdes (approved):

williamdes (approved):

williamdes:
@nicolaasuni could you review this one please ?

nicolaasuni:
The producer line is intentionally hardcoded.
New features like this one could be considered for the new tc-lib-pdf project.

PR #575, opened 2022-12-21
Check fonts files with phpstan too

mvorisek:
https://github.com/phpstan/phpstan/issues/7955 was fixed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -132,4 +132,4 @@ jobs:
             - name: Install phpstan
               run: composer require --dev phpstan/phpstan
             - name: Analyse files
-              run: ./vendor/bin/phpstan --memory-limit=2G
+              run: ./vendor/bin/phpstan --memory-limit=6G
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -4,8 +4,6 @@ parameters:
         - ./
     excludePaths:
         - vendor/
-        # remove once https://github.com/phpstan/phpstan/issues/7955 is fixed
-        - fonts/
         - tests/
 
     scanFiles:


williamdes (dismissed):

.github/workflows/tests.yml
```diff
@@ -132,4 +132,4 @@ jobs:
             - name: Install phpstan
               run: composer require --dev phpstan/phpstan
             - name: Analyse files
-              run: ./vendor/bin/phpstan --memory-limit=2G
+              run: ./vendor/bin/phpstan --memory-limit=-1
```
williamdes, line 135: Thats kind of evil ^^
Maybe @staabm would like to debug this phpstan memory issue?

staabm: > when parallel processing is enabled (default)

yeah, thats expected. every worker proces has the limit you define here. so e.g. with 4 workers you allow phpstan to consume 4x memory-limit to consume

staabm: btw: I am fine with analyzing stuff, but just running phpstan on this project already shows me that alone analyzing the main file takes 7 minutes.

```
/Users/staabm/workspace/TCPDF/tcpdf.php
--- consumed 390 MB, total 930 MB, took 442.72 s
```

@mvorisek I would have expected that someone with interesst in this TCPDF project would have at least analyzed the slow files or similar. expecting that I do that all for you is like a bad joke.

just telling some random open source developer spending their freetime on your problems is not the way how this works

williamdes: > expecting that I do that all for you is like a bad joke.
just telling some random open source developer spending their freetime on your problems is not the way how this works

I apologize for the ping then, I should have looked at it closer before

staabm: fine then. I think analyzing this project could be way faster by splitting the huge files into parts and getting rid of the manual require/include calls but use regular autoloading instead.

mvorisek: we need to wait for 1.9.5 phpstan release, then let's check the performance and memory requirements again

staabm: fwiw, I did my measure with 1.9.x-dev

williamdes (dismissed):

nicolaasuni (approved):

PR #581, opened 2023-01-15
Fix composite glyph output

ren1244:
Fix Composite Glyph Output

overview

## Problem Reproduction

1. Test Font: [kaiu.ttf](https://github.com/thumbb13555/PDFMakerExample/blob/master/app/src/main/assets/kaiu.ttf)
2. Test Code
```php
<?php
require('vendor/autoload.php');

// add test font
$font = TCPDF_FONTS::addTTFfont('kaiu.ttf');

// create new PDF document
$pdf = new TCPDF(PDF_PAGE_ORIENTATION, PDF_UNIT, PDF_PAGE_FORMAT, true, 'UTF-8', false);

// remove default header/footer
$pdf->setPrintHeader(false);
$pdf->setPrintFooter(false);

// set default monospaced font
$pdf->SetDefaultMonospacedFont(PDF_FONT_MONOSPACED);

// set margins
$pdf->SetMargins(PDF_MARGIN_LEFT, PDF_MARGIN_TOP, PDF_MARGIN_RIGHT);

// set default font subsetting mode
$pdf->setFontSubsetting(true);

// set font
$pdf->SetFont($font, '', 12);

// add a page
$pdf->AddPage();

// get esternal file content
$utf8text = "月落烏啼霜滿天\n江楓漁火對愁眠\n姑蘇城外寒山寺\n夜半鐘聲到客船";

// write the text
$pdf->Write(5, $utf8text, '', 0, '', false, 0, false, false, 0);

//Close and output PDF document
$pdf->Output('example_008.pdf', 'I');
```

## Related Reference

[https://learn.microsoft.com/en-us/typography/opentype/spec/otff](https://learn.microsoft.com/en-us/typography/opentype/spec/otff)

> All tables must begin on four-byte boundaries, and any remaining space between tables must be padded with zeros. The length of each table should be recorded in the table record with the actual length of data, not the padded length.

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1323,14 +1323,9 @@ class TCPDF_FONTS {
 					// set the checkSumAdjustment to 0
 					$table[$tag]['data'] = substr($table[$tag]['data'], 0, 8)."\x0\x0\x0\x0".substr($table[$tag]['data'], 12);
 				}
-				$pad = 4 - ($table[$tag]['length'] % 4);
-				if ($pad != 4) {
-					// the length of a table must be a multiple of four bytes
-					$table[$tag]['length'] += $pad;
-					$table[$tag]['data'] .= str_repeat("\x0", $pad);
-				}
 				$table[$tag]['offset'] = $offset;
 				$offset += $table[$tag]['length'];
+				$offset = $offset + 3 & ~3;
 				// check sum is not changed (so keep the following line commented)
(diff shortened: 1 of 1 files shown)


ren1244:
This PR also fix a bug about subset.
Please review tcpdf_fonts.php line 1386.
CheckSumAdjustment write wrong position.
It cause unpredictable behavior.

ren1244:
Update code to fit _getTTFtableChecksum requirements

nicolaasuni (approved):

PR #584, opened 2023-01-20
fix for #583

### LINKED ISSUE #583: Trying to access array offset on value of type bool with PDF/A
Hi,

when building a pdf/a i get this exception:

```php
Trying to access array offset on value of type bool

  at vendor/tecnickcom/tcpdf/tcpdf.php:9917
    9913▕                               }
    9914▕                               $out .= ' >> >>';
    9915▕                       }
    9916▕                       $font = $this->getFontBuffer('helvetica');
  ➜ 9917▕                       $out .= ' /DA (/F'.$font['i'].' 0 Tf 0 g)';
    9918▕                       $out .= ' /Q '.(($this->rtl)?'2':'0');
    9919▕                       //$out .= ' /XFA ';
    9920▕                       $out .= ' >>';
    9921▕                       // signatures
```

the problem is IMO in the line 
https://github.com/tecnickcom/TCPDF/blob/2fb1c01bc37487d1f94fe1297f8d8ad1b5c290bb/tcpdf.php#L9928

which should be
```php
 $font = $this->getFontBuffer((($this->pdfa_mode) ? 'pdfa' : '') .'helvetica');
 ```

can you reproduce and fix, thanks!

### PR DESCRIPTION
**Pull request by simonbuehler**
should be ok now, fixes #583

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -9925,7 +9925,7 @@ class TCPDF {
 				}
 				$out .= ' >> >>';
 			}
-			$this->getFontBuffer((($this->pdfa_mode) ? 'pdfa' : '') .'helvetica');
+			$font = $this->getFontBuffer((($this->pdfa_mode) ? 'pdfa' : '') .'helvetica');
 			$out .= ' /DA (/F'.$font['i'].' 0 Tf 0 g)';
 			$out .= ' /Q '.(($this->rtl)?'2':'0');
 			//$out .= ' /XFA ';
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -9925,7 +9925,7 @@ class TCPDF {
 				}
 				$out .= ' >> >>';
 			}
-			$font = $this->getFontBuffer('helvetica');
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tcpdf.php [hunk 1]
```diff
@@ -9925,7 +9925,7 @@ protected function _putcatalog() {
 				}
 				$out .= ' >> >>';
 			}
-			$font = $this->getFontBuffer('helvetica');
+			$font = $this->getFontBuffer((($this->pdfa_mode) ? 'pdfa' : '') .'helvetica');
```
williamdes, line 9928: So there is a pdfahelvetica mode?

Can you create a new example in the examples folder?
Or update one that is for pdfa?

This will be run as a test on our test suite

simonbuehler: the sample 065 just needs a 

```php
$certificate = 'file:///my.pem';
$privateKey  = 'file://my.prv.key.pem';
// set document signature
$pdf->setSignature($certificate, $privateKey, '', '', 1, array());
```
for the error to pop up

nicolaasuni (approved):

PR #588, opened 2023-02-03
Fix PHP8.2 str_split function returns empty arrays for empty strings

HuongNV13:
In PHP 8.2, the str_split function will returns empty arrays for empty strings.
See: https://php.watch/versions/8.2/str_split-empty-string-empty-array

We can use mb_str_split() instead

### DIFF
--- a/include/barcodes/qrcode.php
+++ b/include/barcodes/qrcode.php
@@ -638,7 +638,8 @@ class QRcode {
 		$barcode_array['bcode'] = array();
 		foreach ($qrTab as $line) {
 			$arrAdd = array();
-			foreach (str_split($line) as $char) {
+            $chars = function_exists('mb_str_split') ? mb_str_split($line) : str_split($line);
+			foreach ($chars as $char) {
 				$arrAdd[] = ($char=='1')?1:0;
 			}
 			$barcode_array['bcode'][] = $arrAdd;
@@ -1307,7 +1308,8 @@ class QRcode {
 				return $this->eatAn();
 			}
 		}
-		$this->items = $this->appendNewInputItem($this->items, QR_MODE_NM, $run, str_split($this->dataStr));
+        $data = function_exists('mb_str_split') ? mb_str_split($this->dataStr) : str_split($this->dataStr);
+		$this->items = $this->appendNewInputItem($this->items, QR_MODE_NM, $run, $data);
 		return $run;
 	}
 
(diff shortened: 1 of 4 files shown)


williamdes (dismissed):

fooman:
Unfortunately [mb_str_split()](https://www.php.net/manual/en/function.mb-str-split.php) was only introduced in php7.4 which will make swapping this out trickier.

williamdes:
> Unfortunately [mb_str_split()](https://www.php.net/manual/en/function.mb-str-split.php) was only introduced in php7.4 which will make swapping this out trickier.

Thanks for the hint!
Okay, some code needs to be added and use PHP_MAJOR_VERSION const and it's friends

williamdes:
or maybe just `function_exists()`

HuongNV13:
Thanks, everyone.
I have updated the code to use function_exists() to use the correct method.

williamdes (approved):

MacGritsch:
I think this is not a valid fix as mb_str_split() behaves the same way as str_split() in PHP 8.2 and above.
So why dont check if its an empty string (and if its needed in that case).
Otherwise you could also write a function str_split_old() which returns array('') if the input string is empty.

include/barcodes/qrcode.php
```diff
@@ -638,7 +638,8 @@ public function __construct($code, $eclevel = 'L') {
 		$barcode_array['bcode'] = array();
 		foreach ($qrTab as $line) {
 			$arrAdd = array();
-			foreach (str_split($line) as $char) {
+            $chars = function_exists('mb_str_split') ? mb_str_split($line) : str_split($line);
```
survik1, line 641: Looks like incorrect indentation, project uses tabs while this code introduces spaces. Same can be said about all other new lines in this MR.

nicolaasuni:
Could you please instead create a wrapper function in tcpdf_static.php?

HuongNV13:
Hi Nicola,
Sorry for the late reply, I was busy with my project.
I will update the patch soon.

Thanks,

nicolaasuni:
Hi,
If this is still an issue please try to create a wrapper function in tcpdf_static.php.
Alternatively please consider using the new tc-lib-pdf if you don't need HTML/CSS/SVG/Javascript features that are still in development.

williamdes:
> If this is still an issue please try to create a wrapper function in tcpdf_static.php.

Very good idead
@HuongNV13 can you do it ?

This PR should be marked as draft for now

nicolaasuni:
Thank you very much, @HuongNV13, for tracking down the PHP 8.2 `str_split()` behaviour change and proposing a fix - it is genuinely appreciated.

The empty-string edge case you identified is handled correctly in the modern PHP string utilities used by **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)**. If you encounter a related issue there, feel free to open a PR or issue in that repository. If you need any help adapting your fix to the new library, please open a discussion and I will do my best to assist.

PR #594, opened 2023-03-07
show correct error on incorrect encoding

rubo77:
Signed-off-by: Ruben Barkow-Kuder <github@r.z11.de>

E.G if you try this:

    $pdf->multiCell(0, 4, 'Some non-UTF-8 text: '.unhtmlentities('&uuml;'), 0, 'L');

You will get this error on PHP 8.1:

    Fatal error: Uncaught ValueError: max(): Argument #1 ($value) must contain at least one element in tcpdf.php:6412

This commit will show a useful error with the causing string

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -2002,7 +2002,11 @@ class TCPDF_FONTS {
 		if ($isunicode) {
 			// requires PCRE unicode support turned on
 			$chars = TCPDF_STATIC::pregSplit('//','u', $str, -1, PREG_SPLIT_NO_EMPTY);
-			$carr = array_map(array('TCPDF_FONTS', 'uniord'), $chars);
+			if(empty($chars)){
+				die('TCPDF Error: Incorrect encoding: '.$str);
+			}else{
+				$carr = array_map(array('TCPDF_FONTS', 'uniord'), $chars);
+			}
 		} else {
 			$chars = str_split($str);
 			$carr = array_map('ord', $chars);


williamdes (changes requested):
This does not feel like the right fix 🤔

rubo77:
The problem is, that this line:
    // get array of unicode values
    $chars = TCPDF_FONTS::UTF8StringToArray($s, $this->isunicode, $this->CurrentFont);
	
sets $char to an empty array if $s contains non-utf8 characters

rubo77:
I moved the error code generation into the function `UTF8StringToArray()`.

The problem always was before PHP8: if there are non-utf8- characters, the PDF generation was just ignoring this whole Cell, leaving it empty without warning.

This should enlight users a bit, what is going wrong then

rubo77:
Maybe we should add `and !empty($str)` ?

include/tcpdf_fonts.php
```diff
@@ -2002,7 +2002,11 @@ public static function UTF8StringToArray($str, $isunicode, &$currentfont) {
 		if ($isunicode) {
 			// requires PCRE unicode support turned on
 			$chars = TCPDF_STATIC::pregSplit('//','u', $str, -1, PREG_SPLIT_NO_EMPTY);
-			$carr = array_map(array('TCPDF_FONTS', 'uniord'), $chars);
+			if(empty($chars)){
+				die('TCPDF Error: Incorrect encoding: '.$str);
+			}else{
+				$carr = array_map(array('TCPDF_FONTS', 'uniord'), $chars);
```
williamdes, line 2008: please change this for PHP 8.2 compat, see #633 (#632)

rubo77:
Please resolve this someone else. Sorry I don't have the time any more

PR #595, opened 2023-03-16
Fix PHP Warning: chr() expects parameter 1 to be int, string given

rubo77:
in include/tcpdf_fonts.php on line 1674

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1671,7 +1671,11 @@ class TCPDF_FONTS {
 			return chr($c);
 		} elseif ($c <= 0x7F) {
 			// one byte
-			return chr($c);
+			if (is_int($c)) {
+				return chr($c);
+			} else {
+				return chr(intval($c));
+			}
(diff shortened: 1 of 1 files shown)


include/tcpdf_fonts.php
```diff
@@ -1671,7 +1671,11 @@ public static function unichr($c, $unicode=true) {
 			return chr($c);
 		} elseif ($c <= 0x7F) {
 			// one byte
-			return chr($c);
+			if (is_int($c)) {
+				return chr($c);
+			} else {
+				return chr(intval($c));
```
williamdes, line 1677: that's already done some lines above, `$c = intval($c);`

rubo77:
+1

I already wondered, if such an old bug isn't fixed otherwise already

williamdes:
Are you using the latest TCPDF ?

The fix was made in https://github.com/rubo77/TCPDF/commit/e2deae00e5ac436560667d6afe1fe2bac3ec8859

Released since: 6.3.0

PR #598, opened 2023-03-19
Typehints for getHeaderMargin() and setHeaderMargin() are inconsistent

fisharebest:
This PR just updates a couple of inconsistent type-hints in the phpdoc.

The header/footer margins are both `float`, but the getters/setters have  inconsistent PHPDOC.

`getFooterMargin()` has `@return float`, but `setFooterMargin()` has `@param int $fm`.

`getHeaderMargin()` has `@return float`, but `setHeaderMargin()` has `@param int $hm`.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -574,12 +574,14 @@ class TCPDF {
 	/**
 	 * Minimum distance between header and top page margin.
 	 * @protected
+	 * @var float
 	 */
 	protected $header_margin;
 
 	/**
 	 * Minimum distance between footer and bottom page margin.
 	 * @protected
+	 * @var float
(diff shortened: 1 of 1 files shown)


williamdes (approved):
That's okay, but could you also add `@var float` on the properties?

fisharebest:
> that's okay, but could you also add @var float on the properties?

Sure!

I wasn't sure of the project's policy - so I just changed the external interfaces.

williamdes (approved):
Thank you for this PR
Now we need to be very patient until it can be merged ;)

nicolaasuni (approved):

PR #601, opened 2023-03-24
 Fix: #553

### LINKED ISSUE #553: GD library error: imagecreatefrompng(): gd-png: libpng warning: iCCP: known incorrect sRGB profile
HI,

I have facing an issues when upload the image to be display for pdf using path.
 after I have amend the tcpdf.php and change as below :

$img = imagecreatefrompng($file);  ===> $img = @imagecreatefrompng($file);

then the files is working fine. But currently i try using imageExpression to pass the value to the path, but the image not working.

Anyone can assists?

Thanks.

lpetrora:
Fix: #553

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -7121,6 +7121,11 @@ class TCPDF {
 			$mtd = '_parse'.$type;
 			// GD image handler function
 			$gdfunction = 'imagecreatefrom'.$type;
+			
+			if ($type === 'png') {
+				$gdfunction = '@' . $gdfunction;
+			}
+
 			$info = false;
 			if ((method_exists('TCPDF_IMAGES', $mtd)) AND (!($resize AND (function_exists($gdfunction) OR extension_loaded('imagick'))))) {
 				// TCPDF image functions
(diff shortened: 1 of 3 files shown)


lpetrora:
I have already signed the CLA but the status is still pending. I tried to recheck but there is no change

williamdes:
Please edit the description to use `Fixes: #553` or `Fix: #553`

tcpdf.php
```diff
@@ -7121,6 +7121,11 @@ public function Image($file, $x=null, $y=null, $w=0, $h=0, $type='', $link='', $
 			$mtd = '_parse'.$type;
 			// GD image handler function
 			$gdfunction = 'imagecreatefrom'.$type;
+			
+			if ($type === 'png') {
+				$gdfunction = '@' . $gdfunction;
```
mvorisek, line 7126: this cannot work - https://3v4l.org/lbu9o

williamdes (changes requested):
Thanks for the testing @mvorisek 

This PR does not work correctly: https://github.com/tecnickcom/TCPDF/pull/601#discussion_r1174422155

THenkeDE:
to bypass this you have to add a helper variable like

```
			$gdfunction = 'imagecreatefrom'.$type;
			$silenceGdfunction = $type === 'png';
			...
					// GD library
					$img = $silenceGdfunction ? @$gdfunction($file) : $gdfunction($file);
```
but i have no idea if silencing a method was, is oder ever will be a good decision.

nicolaasuni:
It looks like this PR is not yet ready as there are unsolved comments.
I agree that suppressing errors is not actually a fix.
NOTE: the new tc-lib-pdf library has already a better image support if you can wait for the other HTML/CSS/SVG/Javascript features to be ported.

Issue #577, opened 2022-12-25
tcpdf_import


vinojz:
This file has not been updated for a long time，When do you plan to update it.

williamdes:
See #574
I'd say never ^^

Issue #574, opened 2022-12-17
Delete non used and non usable import code? 


williamdes:
Hi @nicolaasuni

Could we delete the tcpdf import code as it is not finished and should not be used by anyone?

That would reduce this library a bit

nagyimre1980:
files to delete:
tcpdf_import.php
tcpdf_parser.php
include/tcpdf_filters.php

williamdes:
Yes, @nicolaasuni please let us know

vinojz:
I have a PDF file, how can I show it in the browser.

williamdes:
> I have a PDF file, how can I show it in the browser.

Off topic, sorry

williamdes:
@nicolaasuni opinion needed on this one. Else close it

jansor:
hi Just wanne check-up on this, i dont know the code well enough, but effectively we use [https://github.com/Setasign/FPDI](https://github.com/Setasign/FPDI) to import pdf files, i am not sure if they use any direct part of the files mentioned above? would this break Setasign/FPDI ?? Just concerned.

Thank you for the continued work on this

williamdes:
> hi Just wanne check-up on this, i dont know the code well enough, but effectively we use https://github.com/Setasign/FPDI to import pdf files, i am not sure if they use any direct part of the files mentioned above? would this break Setasign/FPDI ?? Just concerned.

99.9% chance they are unrelated. The code at TCPDF for importing is reported as unfinished

PR #605, opened 2023-04-05
feat: remove examples from packaging

shyim:
image
saves 2MB in everyones vendor folder :)


williamdes (approved):
Well, that will not help me at Debian packaging, but I will manage to workaround it
Thanks !

shyim:
composer package as debian package? 😱

williamdes:
> composer package as debian package? scream

Well, no. But in a sense yes.
Let me explain, the file you just changed controls the command `git archive`. GitHub uses that to give you nice tarballs on branches and tags/releases.

The tarballs are re-used by packagist/composer. And also packaging people in Debian or other distributions.
See: https://salsa.debian.org/phpmyadmin-team/tcpdf
See: https://salsa.debian.org/phpmyadmin-team/tcpdf/-/blob/debian/6.6.2+dfsg1-1_bpo11+1/debian/watch (the file used by [Debian watch](https://wiki.debian.org/debian/watch) )

So if you exclude tests or other files from the git archive, other users who need them will have to use git clone to have them back. It's not very cool but sometimes it's best for everyone to have them excluded from the git archive.
Because at Debian the tests will be run on the source code to ensure the packaged code really works: https://salsa.debian.org/phpmyadmin-team/tcpdf/-/blob/debian/6.6.2+dfsg1-1_bpo11+1/debian/tests/control

Does that help understand ?

shyim:
Ahhh okay. Thanks for maintaining phpmyadmin on debian! ❤️

williamdes:
> Ahhh okay. Thanks for maintaining phpmyadmin on debian! :heart:

And more of them ;p
I wrote an article right now: https://blog.williamdes.eu/Debian/the-debian-packages-i-maintain/

williamdes (approved):

nicolaasuni:
@williamdes I will look into sign the tags.

nicolaasuni:
@williamdes Is this something worth doing or shall we focus on the new tc-lib-pdf lib instead?

williamdes:
> @williamdes Is this something worth doing or shall we focus on the new tc-lib-pdf lib instead?

Removing the examples will save quite some bytes on each server that downloads tcpdf.
Merging this would be great.

Thank you for looking into git tags and commits signing.

I would argue that we could also drop https://github.com/tecnickcom/TCPDF/blob/main/tcpdf_import.php and https://github.com/tecnickcom/TCPDF/blob/main/tcpdf_parser.php
Conversation about this on https://github.com/tecnickcom/TCPDF/issues/574 and https://github.com/tecnickcom/TCPDF/issues/577 (duplicate)

williamdes:
@nicolaasuni what do you think about this PR and my last message ?

PR #611, opened 2023-05-04
Fixed problems with S25 barcode

theonlytruth:
Using your S25 barcode generator I found a problem. The representation of the digit 8 was the same as 0. I fixed it. Now it is recognized correctly.

Additionally you were using different bar widths for the start and stop values. I changed this too.

One thing I did not change till now is the part where you add an leading 0 if the string is odd. Why do you do so? I can't find anything that S25 has to be even. Maybe you mixed it up with interleaved 2 of 5 bar code where you need the even number because the digits are represented from pairs of bars and whitespaces?

### DIFF
--- a/tcpdf_barcodes_1d.php
+++ b/tcpdf_barcodes_1d.php
@@ -828,7 +828,7 @@ class TCPDFBarcode {
 		$chr['5'] = '11101011101010';
 		$chr['6'] = '10111011101010';
 		$chr['7'] = '10101011101110';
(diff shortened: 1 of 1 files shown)


nicolaasuni:
Can you please check/use the https://github.com/tecnickcom/tc-lib-barcode instead?

nicolaasuni (approved):
This was already fixed in tc-lib-barcode.

PR #613, opened 2023-05-09
Fix return type annotation

survik1:
A small annotation fix with in line with PHPStan conditional returns types introduced in 1.6 

https://phpstan.org/blog/phpstan-1-6-0-with-conditional-return-types

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -4101,7 +4101,7 @@ class TCPDF {
 	 * @param string $fontstyle Font style. Possible values are (case insensitive):<ul><li>empty string: regular</li><li>B: bold</li><li>I: italic</li><li>U: underline</li><li>D: line-through</li><li>O: overline</li></ul> or any combination. The default value is regular.
 	 * @param float $fontsize Font size in points. The default value is the current size.
 	 * @param boolean $getarray if true returns an array of characters widths, if false returns the total length.
-	 * @return float[]|float total string length or array of characted widths
+	 * @return ($getarray is true ? float[] : float) total string length or array of characted widths
 	 * @author Nicola Asuni
 	 * @public
 	 * @since 1.2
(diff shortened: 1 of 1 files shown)


tcpdf.php
```diff
@@ -4101,7 +4101,7 @@ public function setTextColor($col1=0, $col2=-1, $col3=-1, $col4=-1, $ret=false,
 	 * @param string $fontstyle Font style. Possible values are (case insensitive):<ul><li>empty string: regular</li><li>B: bold</li><li>I: italic</li><li>U: underline</li><li>D: line-through</li><li>O: overline</li></ul> or any combination. The default value is regular.
 	 * @param float $fontsize Font size in points. The default value is the current size.
 	 * @param boolean $getarray if true returns an array of characters widths, if false returns the total length.
-	 * @return float[]|float total string length or array of characted widths
+	 * @return ($getarray is true ? float[] : float) total string length or array of characted widths
```
williamdes, line 4104:
```suggestion
	 * @return float[]|float total string length or array of characted widths
	 * @phpstan-return ($getarray is true ? float[] : float) total string length or array of characted widths
```
Let's keep backwards compat for other IDEs or tools

tcpdf.php
```diff
@@ -4117,7 +4117,7 @@ public function GetStringWidth($s, $fontname='', $fontstyle='', $fontsize=0, $ge
 	 * @param string $fontstyle Font style. Possible values are (case insensitive):<ul><li>empty string: regular</li><li>B: bold</li><li>I: italic</li><li>U: underline</li><li>D: line through</li><li>O: overline</li></ul> or any combination. The default value is regular.
 	 * @param float $fontsize Font size in points. The default value is the current size.
 	 * @param boolean $getarray if true returns an array of characters widths, if false returns the total length.
-	 * @return float[]|float total string length or array of characted widths
+	 * @return ($getarray is true ? float[] : float) total string length or array of characted widths
```
williamdes, line 4120:
```suggestion
	 * @return float[]|float total string length or array of characted widths
	 * @phpstan-return ($getarray is true ? float[] : float) total string length or array of characted widths
```

tcpdf.php
survik1: Sure

williamdes (approved):
Awesome

PR #627, opened 2023-06-07
Fix non-numeric value warning

d-salerno:
Fixes this warning on generating PDF:
Warning: A non-numeric value encountered in /tcpdf/tcpdf.php on line 5473

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -5467,6 +5467,9 @@ class TCPDF {
 				$s .= 'q '.$this->TextColor.' ';
 			}
 			// rendering mode
+			if(!is_numeric($this->textstrokewidth)){
+ 				$this->textstrokewidth = 0;
+			}
 			$s .= sprintf('BT %d Tr %F w ET ', $this->textrendermode, ($this->textstrokewidth * $this->k));
 			// count number of spaces
 			$ns = substr_count($txt, chr(32));


tcpdf.php
```diff
@@ -5467,6 +5467,9 @@ protected function getCellCode($w, $h=0, $txt='', $border=0, $ln=0, $align='', $
 				$s .= 'q '.$this->TextColor.' ';
 			}
 			// rendering mode
+			if(!is_numeric($this->textstrokewidth)){
+ 				$this->textstrokewidth = 0;
+			}
```
williamdes, line 5472: Can you try to track in the code where it changes to a non numeric value ?

```php
	/**
	 * Text stroke width in doc units.
	 * @protected
	 * @since 4.9.008 (2010-04-03)
	 */
	protected $textstrokewidth = 0;
```
Since it's initialized to 0
Maybe you can add an `@var int` or `@var float` or `@var int|float` phpdoc on this property, for documentation purposes


The fix you propose indeed should work, but if we can catch it earlier to avoid the `textstrokewidth` being a non numeric that would be cool

d-salerno: OK, turns out I was passing a non-numeric value to the `Text` function for `$fstroke`. The new commit ignores non-numeric values that are passed

tcpdf.php
```diff
@@ -22055,7 +22055,7 @@ public function getNumberOfColumns() {
 	public function setTextRenderingMode($stroke=0, $fill=true, $clip=false) {
 		// Ref.: PDF 32000-1:2008 - 9.3.6 Text Rendering Mode
 		// convert text rendering parameters
-		if ($stroke < 0) {
+		if ($stroke < 0 or !is_numeric($stroke)) {
```
williamdes, line 22058:
```suggestion
		if ($stroke < 0 || !is_numeric($stroke)) {
```

tcpdf.php
williamdes: Thanks, it looks way better

Is this mistake related to your code of tcpdf code calling the function ?

d-salerno: I guess so, but anybody calling `$pdf->Text($x, $y, $text, "non-numeric");` would get the same warning

nicolaasuni (approved):

PR #630, opened 2023-06-29
php 8+ edge cases fixes

bruno-farias:
This PR fixes issues with division by zero, returns from function pregSplit to match the expected, and calculates maximum width for a single character on the string will not throw an error when the variable is not countable

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -6409,7 +6409,7 @@ class TCPDF {
 		// calculate maximum width for a single character on string
 		$chrw = $this->GetArrStringWidth($chars, '', '', 0, true);
 		array_walk($chrw, array($this, 'getRawCharWidth'));
-		$maxchwidth = max($chrw);
+		$maxchwidth = ((is_array($chrw) || $chrw instanceof Countable) && count($chrw) > 0) ? max($chrw) : 0;
 		// get array of chars
 		$uchars = TCPDF_FONTS::UTF8ArrayToUniArray($chars, $this->isunicode);
 		// get the number of characters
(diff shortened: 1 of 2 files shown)


williamdes (approved):
Thank you !

edoniti:
@williamdes I am having a problem that is solved from this pull request (max function throwing an error because of the empty array). Can we except this to be merged to the main branch?

edoniti (approved):

williamdes:
Hi @nicolaasuni
Would you mind giving a review of pending PRs?

tcpdf.php
```diff
@@ -6409,7 +6409,7 @@ public function Write($h, $txt, $link='', $fill=false, $align='', $ln=false, $st
 		// calculate maximum width for a single character on string
 		$chrw = $this->GetArrStringWidth($chars, '', '', 0, true);
 		array_walk($chrw, array($this, 'getRawCharWidth'));
-		$maxchwidth = max($chrw);
+		$maxchwidth = ((is_array($chrw) || $chrw instanceof Countable) && count($chrw) > 0) ? max($chrw) : 0;
```
kallesommernielsen, line 6412: If `$chrw` is not an `array`, then the above `array_walk()` would also cause a warning. So it seems this should be something along the lines of:

```php
$maxchwidth = 0;

if (is_array($chrw) && count($chrw) > 0) {
    array_walk($chrw, array($this, 'getRawCharWidth'));
    $maxchwidth = max($chrw);
}
```

I did not see any places where `GetArrStringWidth()` could return `object`, so I omitted the `instanceof` check

nicolaasuni (approved):

PR #633, opened 2023-08-01
Fixes for PHP 8.2 in tcpdf_fonts.php - fixes #632

### LINKED ISSUE #632: Use of "self" in callables is deprecated in include/tcpdf_fonts.php (PHP 8.2)
An issue with a suggested fix. A "use of "self" in callables is deprecated" warning is arising from tcpdf_fonts.php using PHP 8.2

In lines 1788, 1790, and 2010 of tcpdf_fonts.php, this deprecation warning (PHP 8.2) can be fixed by changing the lines to read:

`return array_map(self::class . "::unichrUnicode",$ta);`

`return array_map(self::class . "::unichrASCII",$ta);` and

`$carr = array_map(self::class . "::uniord",$chars);` respectively.

sagehenstudio:
Fixes "use of "self" in callables is deprecated" warning is arising from tcpdf_fonts.php when using PHP >= 8.2

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1780,9 +1780,9 @@ class TCPDF_FONTS {
 	 */
 	public static function UTF8ArrayToUniArray($ta, $isunicode=true) {
 		if ($isunicode) {
-			return array_map(self::class . "::unichrUnicode",$ta);
+			return array_map(array('self','unichrUnicode'),$ta);
 		}
-		return array_map(self::class . "::unichrASCII",$ta);
+		return array_map(array('self','unichrUnicode'),$ta);
 	}
 
 	/**
@@ -2002,7 +2002,7 @@ class TCPDF_FONTS {
 		if ($isunicode) {
 			// requires PCRE unicode support turned on
 			$chars = TCPDF_STATIC::pregSplit('//','u', $str, -1, PREG_SPLIT_NO_EMPTY);
-			$carr = array_map(self::class . "::uniord",$chars);
+			$carr = array_map(array('self','uniord'),$chars);
 		} else {
 			$chars = str_split($str);
 			$carr = array_map('ord', $chars);
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1780,9 +1780,9 @@ class TCPDF_FONTS {
 	 */
 	public static function UTF8ArrayToUniArray($ta, $isunicode=true) {
 		if ($isunicode) {
-			return array_map(array('self','unichrUnicode'),$ta);
+			return array_map(array('self','unichrUnicode'), $ta);
 		}
-		return array_map(array('self','unichrUnicode'),$ta);
+		return array_map(array('self','unichrUnicode'), $ta);
 	}
 
 	/**
@@ -2002,7 +2002,7 @@ class TCPDF_FONTS {
 		if ($isunicode) {
 			// requires PCRE unicode support turned on
 			$chars = TCPDF_STATIC::pregSplit('//','u', $str, -1, PREG_SPLIT_NO_EMPTY);
-			$carr = array_map(array('self','uniord'),$chars);
+			$carr = array_map(array('self','uniord'), $chars);
 		} else {
 			$chars = str_split($str);
 			$carr = array_map('ord', $chars);
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1780,9 +1780,9 @@ class TCPDF_FONTS {
 	 */
 	public static function UTF8ArrayToUniArray($ta, $isunicode=true) {
 		if ($isunicode) {
-			return array_map(array('TCPDF_FONTS', 'unichrUnicode'), $ta);
+			return array_map(self::class . "::unichrUnicode",$ta);
 		}
-		return array_map(array('TCPDF_FONTS', 'unichrASCII'), $ta);
+		return array_map(self::class . "::unichrASCII",$ta);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Do you know if it breaks old PHP versions ?
We still support 5.4..

sagehenstudio:
PHP 5.4... Wow. This should work down to PHP 5.5.

I guess I'd have to write some sort of phpversion() check. Would that be OK?

williamdes:
> PHP 5.4... Wow. This should work down to PHP 5.5.
> 
> I guess I'd have to write some sort of phpversion() check. Would that be OK?

I checked composer.json it's even worse: 5.3
I think that would be okay, keep your patch for all versions and do an if for old versions to have the old version of the code

sagehenstudio:
5.3! Gasp! 😋 I was able to test and confirm another approach (committed to #633) works using PHP 5.3.29 through 8.2.

williamdes (approved):
Awesome!!!
You removed one space before the argument, can you add it back?

sagehenstudio:
Spaces done, I think?

williamdes (approved):
Yup look great

williamdes:
Please add `Fixes: #632` to your PR description so it links it and closes it on merge

sagehenstudio:
I've made a mistake. This fix doesn't work, and probably won't work without a phpversion() conditional and an include() to move code (self::class) which WILL break 5.3 to another file. Hopefully other ideas roll in.

sagehenstudio:
This fix starts working with PHP 5.5.

It seems it'd be a bit of an ordeal to make it work 5.3-8.2 because of these callables. Do-able but not pretty!

sagehenstudio:
Interestingly, PHP 5.3 EOL was almost exactly ten years ago: 14 Aug 2014 😳

williamdes:
> Interestingly, PHP 5.3 EOL was almost exactly ten years ago: 14 Aug 2014 😳

😳😨

Can it be a variable and replaced if version compare === old?

sagehenstudio:
5.3 parses everything before run and when it sees self::class, it breaks. ☹️ That's why I was mentioning using phpversion() to check for >8.2 and including that in a separate file, maybe.

williamdes:
> self::class

If I am not mistaken `self::class` is just a string in the end, can you hard code the string ?
I would bet it's just `\CLASS_NAME`

sagehenstudio:
**self::class** is a string but 5.3 sees that :: in the file and 😵

Something like this might work:

`$selfclass = get_called_class();
if ( $isunicode ) {
    return array_map($selfclass . '::unichrUnicode', $ta );
}
return array_map($selfclass . '::unichrASCII', $ta );`

I'm testing now.

sagehenstudio:
Sure, TCPDF_Fonts. Have you tried what you're suggesting?

williamdes:
> Sure, TCPDF_Fonts. Have you tried what you're suggesting?

Hmm, sorry I got confused in my brain

So with a version compare it can be changed between `TCPDF_FONTS` and `self`.
Should be okay for the linter and the code execution, right ?

sagehenstudio:
Because writing self::class _anywhere_ in a file called by TCPDF (such as tcpdf_fonts.php) will break PHP 5.3. Similar to how using array() notation [] in a PHP 5.3 file will break it. The other solution I've mentioned with include() is much more complicated, and involves checking for PHP > 8.2, and including a separate file conditionally which isn't seen by 5.3 at all.

```
if ( version_compare( phpversion(), '5.3', '>' ) ) {
    if ( in_array( 'nuts', [ 'nutty', 'array' ] ) ) {
        // Doesn't even get here due to syntax error
	die( 'Well, nuts' );
    }
}
```

Using **get_called_class()** (like I did above) is the simplest solution I've found that works. I'll do some more testing here to make sure.

williamdes:
> Because writing self::class _anywhere_ in a file called by TCPDF (such as tcpdf_fonts.php) will break PHP 5.3. Similar to how using array() notation [] in a PHP 5.3 file will break it. The other solution I've mentioned with include() is much more complicated, and involves checking for PHP > 8.2, and including a separate file conditionally which isn't seen by 5.3 at all.
> 
> ```
> if ( version_compare( phpversion(), '5.3', '>' ) ) {
>     if ( in_array( 'nuts', [ 'nutty', 'array' ] ) ) {
>         // Doesn't even get here due to syntax error
> 	die( 'Well, nuts' );
>     }
> }
> ```
> 
> Using **get_called_class()** (like I did above) is the simplest solution I've found that works. I'll do some more testing here to make sure.

I agree, that said you do not need to write self::class but only it's final form and it will work great too
that said I like your get_called_class solution, it may solve this nicely without too much lines
it it maybe works for all php versions

williamdes (approved):
Thank you so much for your hard work !

sagehenstudio:
You are welcome!

I've been using this library since well... PHP 5.3, so it's the least I can do to finally jump in on Github and help. 👵🏼

williamdes:
> You are welcome!
> 
> I've been using this library since well... PHP 5.3, so it's the least I can do to finally jump in on Github and help. 👵🏼

So cool!
I will gladly welcome you on other open source projects I contribute to

nicolaasuni (approved):

jmv69800:
Hi
I'm very new in TCPDF.
Also very new in GitHub (first use, in fact). Please, accept my appolgies if I don't use GitHub correctly.

I tested the proposed fixe under **PHP 8.2.9 x86** (in preparation to use TCPDF in PHP 8.3)
... and it seems not working.

I've reported the 3 new lines in tcpdf_fonts.php, and removed the old 3 existing ones, but I obtain that

_**Deprecated: Optional parameter $isunicode declared before required parameter $currentfont is implicitly treated as a required parameter in .\tcpdf\include\tcpdf_fonts.php on line 2000, and 2027, 2043, etc...**_

Also $setbom, $forcertl, $str, $default_css, $tagvs, etc... deprecated (perhaps due to the first deprecated)

Is there something I've done wrong ?

Issue #634, opened 2023-08-02
Next TCPDF release ?


williamdes:
Hi @nicolaasuni here is a summary of fixes and issues addressed. I intentionally kept some other wrong or unfinished ones from this list.

### PHP 8 fixes

- #632 is fixed by #633
- #630
- #615 fixes #614

### PHP 7 fixes

- #620

### Complex fixes

- #581
- #584

### Nice and small footprint features

- #572

### Important LICENSE fix

- #591

### CI and analysis stuff

- #575
- #613
- #598

### Docs

- #589
- #643 

### CI

- #642 

### Barcode fixes

- #611
- #588 (indentation is broken)

### Workarounds
- #627 (not sure we should accept it, since the function parameter has a type and if it's not respected should we workaround it ?)

Hr0bar:
Thanks for your and all other contributors work keeping this up-to-date with latest PHP versions! Much appreciated by our non-profit NGO where we use it to generate documents.

survik1:
Hi, any chance to get new release? We would especially like to silence all the deprecation/notice errors we get from this library.

williamdes:
> Hi, any chance to get new release? We would especially like to silence all the deprecation/notice errors we get from this library.

What merged bug fixes are them?

survik1:
> > Hi, any chance to get new release? We would especially like to silence all the deprecation/notice errors we get from this library.
> 
> What merged bug fixes are them?

Especially everything under "PHP 8 fixes" would be nice.

williamdes:
6.6.3 is out! (changelog: #643 )

PR #589, opened 2023-02-03
README: tone down the warning about tc-lib-pdf

rubo77:

PR #653, opened 2023-10-09
Fixes: missing functionality for svg <use> elements 

### LINKED ISSUE #652: missing functionality for svg <use> elements
With the `ImageSVG` functionality,
[image](https://www.w3.org/Graphics/SVG/)
is not displayed correctly.
image

**location of missing functionality** `startSVGElementHandler` and `endSVGElementHandler`.

**behavior in version 6.6.5**
- objects outside `<defs>` elements can not be referenced with `<use>` since they are not stored in array `$svgdefs`.
- `$svgdefs` does not contain child elements with id attributes set.

**additional info**
- currently the structure does not allow for elements to be stored as child element and independently being referenced. 
- a `<use>` element within a recursive call from a `<use>` can cause issues when addressing the points above.

ghost:
Fixes:  #652


tcpdf.php
```diff
@@ -23944,21 +23958,33 @@ protected function startSVGElementHandler($parser, $name, $attribs, $ctm=array()
 			$this->svgclippaths[$this->svgclipid][] = array('name' => $name, 'attribs' => $attribs, 'tm' => $this->svgcliptm[$this->svgclipid]);
 			return;
 		}
-		if ($this->svgdefsmode AND !in_array($name, array('clipPath', 'linearGradient', 'radialGradient', 'stop'))) {
+		if (!in_array($name, array('clipPath', 'linearGradient', 'radialGradient', 'stop')) AND $this->svgusedepth == 0) {
```
williamdes, line 23961:
```suggestion
		if (!in_array($name, array('clipPath', 'linearGradient', 'radialGradient', 'stop')) AND $this->svgusedepth === 0) {
```

tcpdf.php
```diff
@@ -1641,6 +1641,20 @@ class TCPDF {
 	 */
 	protected $svgdefs = array();
 
+    /**
```
williamdes, line 1644: I am afraid this patch is not very minimal and so will have less changes being accepted into tcpdf

williamdes (approved):
Thank for making this patch !

tcpdf.php
k-nicole: This library is part of the moodle core. We at TU Graz are relying on it to generate certificates, which among other contents contain logos of our partners (SVGs). Its crucial for us that they are correct on the pdf and currently this is not the case. Hence our contribution to this library.
Is there anything that can be done to increase chances? I will change up the contribution in any way necessary to provide this functionality. (:

williamdes: First, thank you for contributing to the libraries you use !

Second, I am here as an advisor and try to monitor issues and pull-requests on my free time
Currently the library owner and maintainer remains mysterious in choices that are made. But it's mostly the rule of more the change is okay and small more it has chances to get on a release one day.

So from time to time I build an issue with a summary of PRs: https://github.com/tecnickcom/TCPDF/issues/634
Maybe it helps, I never know, nobody knows
So it's a wait and see, but you need to get it slim and clean and right. Or it will probably get refused and you are back in a 6-8 months loop to get feedback on the new changes

This recent PR changed the library maintenance official statement, and that's a relief: #589 

In summary, I think this patch will appear too large. Is there other ways to change less code ?

BsAtHome:
Hi,
There are still a few problems with this patch applied. The `<use>` element is more complicated than implemented by this patch. I have a problem file that shows the issue(s) (a cooked down test version from the original shown below).
There are three clipPath scenarios that I tested.
* First is `<use>` in the `<defs><clipPath>` section with a forward reference to a path later on in the document. This scenario fails completely and no clipPath is applied.
* Second is a copied version of the document's `<path>` in the `<defs><clipPath>` section, but it retained the original transformation attribute from the source. This scenario failed to apply the transformation.
* Third is a working version, where the path is normalized.

The used svg (inkscape displays all three version correct):
`<svg id="testUseClippath" width="100%" height="100%" viewBox="0 0 50 30" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">`
`  <defs>`
`    <clipPath id="usepath"><use xlink:href="#realpath" /></clipPath>`
`<!--    <clipPath id="usepath"><path id="realpath" d="M 14,-1 C 10,1 6,9 0,9 C -6,9 -12,1 -14,-1 C -11,-2 -6,-9 0,-9 C 6,-9 11,-2 14,-1 Z" transform="translate(25,15)" /></clipPath> -->`
`<!--    <clipPath id="usepath"><path d="M 39,14 C 35,16 31,24 25,24 C 19,24 13,16 11,14 C 14,13 19,6 25,6 C 31,6 36,13 39,14 Z" /></clipPath> -->`
`  </defs>`
`  <g>`
`    <rect x="0.1" y="0.1" width="49.8" height="29.8" style="fill:none;stroke:#cc0;stroke-width:0.2" />`
`    <g clip-path="url(#usepath)">`
`      <rect x="0" y="0" width="50" height="30" style="fill:#222;stroke:none" />`
`      <ellipse cx="25" cy="25" rx="15" ry="10" style="fill:#800f08;stroke:none" />`
`    </g>`
`    <path id="realpath" d="M 14,-1 C 10,1 6,9 0,9 C -6,9 -12,1 -14,-1 C -11,-2 -6,-9 0,-9 C 6,-9 11,-2 14,-1 Z" transform="translate(25,15)" style="fill:none;stroke:black;stroke-width:1" />`
`  </g>`
`</svg>`

nicolaasuni:

Issue #660, opened 2023-11-03
Automatic calculation of w/h fails for ImageSVG()


d-javu:
Doc states this regarding the w and h parameters: If not specified or equal to zero, it is automatically calculated
However, this does not happen.
The bug appeared in 6.0.031 and was caused by the 'avoid "division by zero"' part of commit: 0c26f23
The fix, for me, is to remove the div_by_zero mitigation.

nicolaasuni:
Thank you for reporting this. TCPDF 6 is now a legacy project and is officially **deprecated**, so no further fixes or features will be merged into this repository. As part of this transition, we are closing older issues. Please migrate to **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** and, if the same problem still occurs there, open a new issue with updated reproduction steps. For more context, see the pinned announcement: [Thank You - and the Future of TCPDF](https://github.com/tecnickcom/TCPDF/issues/867).

PR #661, opened 2023-11-03
ImageSVG() does not calculate the correct w/h values anymore - fix it

d-javu:
Fix for issue #660 Automatic calculation of w/h fails for ImageSVG()
The bug appeared in 6.0.031 and was caused by the
'avoid "division by zero"' part of commit: 0c26f23 The fix, for me, is to remove the div_by_zero mitigation.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -23034,12 +23034,6 @@ class TCPDF {
 				}
 			}
 		}
-		if ($ow <= 0) {
-			$ow = 1;
-		}
-		if ($oh <= 0) {
-			$oh = 1;
-		}
 		// calculate image width and height on document
 		if (($w <= 0) AND ($h <= 0)) {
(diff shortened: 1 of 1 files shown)


williamdes (approved):
okay, let's see what the author thinks about it

nicolaasuni:

d-javu:
Rebased onto your recently force-pushed main branch.

williamdes (approved):

nicolaasuni:
The proposed solution will reintroduce the division by zero issue.
The whole logic should be revisited instead.
I am aiming to have the SVG support in tc-lib-pdf library before the end of the year.

d-javu:
Do you have the problematic file that gives a div by zero error?
If you could make it available, I could take a look at fixing both cases.

PR #671, opened 2023-12-01
fix: base64 images not rendered inside of html

compojoom:
I'm not sure when this happened, but default base64 encoded images were no longer rendered in html  text. 

If the image is rendered like this:
```
<img src="data:image/png;base64,....." />
```

the image was not rendered when inside of an html text. Instead one has to replace `data:image/png;base64,` with `@`which in my opinion is not ideal and an overhead. 

https://github.com/tecnickcom/TCPDF/issues/179

### DIFF
--- a/CHANGELOG.TXT
+++ b/CHANGELOG.TXT
@@ -280,7 +280,7 @@
 	- Bug item #956 "Monospaced fonts are not alignd at the baseline" was fixed.
 	- Bug item #964 "Problem when changing font size" was fixed.
 	- Bug item #969 "ImageSVG with radialGradient problem" was fixed.
-	- sRGB.icc file was replaced with the one from the Debian package icc-profiles-free (2.0.1+dfsg-1) 
+	- sRGB.icc file was replaced with the one from the Debian package icc-profiles-free (2.0.1+dfsg-1)
 
 6.0.091 (2014-08-13)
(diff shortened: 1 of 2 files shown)


williamdes (approved):

nicolaasuni:

williamdes (approved):

compojoom:
@williamdes I rebased. Sorry for the changelog change, but it seems you had an empty space and my IDE wants to remove it :)

williamdes:
> @williamdes I rebased. Sorry for the changelog change, but it seems you had an empty space and my IDE wants to remove it :)

That's okay

But can you remove the other commits?
git reset hard on the base and cherry pick https://github.com/tecnickcom/TCPDF/pull/671/commits/605d0a1ba75f54d9edc9d903213ae0f76b434a02 the force push?

compojoom:
how about now?

williamdes (approved):
perfect

nicolaasuni:
Unfortunately the proposed change has security implications as the file_exist function is used in other places too.
This may allow an attacker to provide arbitrary executable content somewhere.
A proper solution will require a bit of thinking but I'd rather put the energy to complete tc-lib-pdf that already has 90+% of the TCPDF features.

PR #678, opened 2024-01-19
Implemented rectangular data matrix barcode

w512work:
Implemented support for rectangular data matrix code. Corrected values in code definition table and corrected modulo 254 operation in function `get253StateCodeword`.


williamdes:
Do you have an ouput to show here and some example code ?
It would be great to have a new example in the examples/ folder

include/barcodes/datamatrix.php
```diff
@@ -230,38 +237,62 @@ class Datamatrix {
 	 * This is the class constructor.
 	 * Creates a datamatrix object
 	 * @param string $code Code to represent using Datamatrix.
+	 * @param string $shape Shape of datamatrix code (R = rectangular, S = square)
```
williamdes, line 240:
```suggestion
	 * @param string $shape Shape of datamatrix code (R = rectangular, S = square)
	 * @phpstan-param 'R'|'S' $shape
```

williamdes (approved):
The code looks good

include/barcodes/datamatrix.php
w512work: @williamdes Just saw your comment recommendations. Unfortunately only after I added the latest change -> Update will follow...
I've never used phpstan. Is the last line there to specify the parameters for automated testing?

williamdes: You should check out phpstan, it is for static analysis

Here is an example of what my proposal would do https://phpstan.org/r/f04d0c5e-0a40-4959-b03a-fdee4430a29a

w512work: If I understand correctly, this does not generate an automatic verification script.
As you may have seen, I have extended the parameter so that you can specify the desired dimensions (e.g. 48x48). Can this also be specified in a kind of wildcard form, or would each valid combination have to be specified separately?

nicolaasuni:
This feature is already present in https://github.com/tecnickcom/tc-lib-barcode that is already integrated with tc-lib-pdf.
The current datamatrix implementation in TCPDF is not maintained and contains other bugs.
If you want to use datamatrix with TCPDF it is easier to import the tc-lib-barcode and integrate it like in tc-lib-pdf.

include/barcodes/datamatrix.php
williamdes: > If I understand correctly, this does not generate an automatic verification script.

This is used by automatic tools, in this repo:
https://github.com/tecnickcom/TCPDF/blob/5fce932fcee4371865314ab7f6c0d85423c5c7ce/.github/workflows/tests.yml#L121

But for people that will integrate TCPDF, it will also throw warnings if they use phpstan and pass something else than the values you did set

williamdes:
@w512work it is probably best that you migrate to the new library for data matrix and close this PR

w512work:
Thank you @nicolaasuni for the hint.

I do not only need barcodes. If I have understood correctly, tc-lib-pdf is the top level for generating comprehensive PDFs. In its readme it says at the top that the library is not yet ready.
I would have liked to take a look at the examples of this version to get an idea of the current status. I installed the current version of the lib via composer and tried for about an hour without success to get the example script in the examples folder displayed im my browser. I failed because the autoloader could not be loaded. 
I'm not sure if it's a good idea to change everything at this point without causing various other problems that will significantly increase the development and maintenance effort of my project...

Can I view the examples anywhere on the net without spending 2 hours coding, like on https://tcpdf.org/examples/ ?

w512work:
> Can I view the examples anywhere on the net without spending 2 hours coding, like on https://tcpdf.org/examples/ ?

Just tried the command  `make server` mentioned in the readme. Unfortunately it gives me the error:
> Directory example does not exist.
> make: *** [Makefile:250: server] Error 1

The manual approach ` php -t examples -S 10.0.0.251:8888` throws another error when I try to access from the browser:
> [Fri Jan 26 09:55:16 2024] PHP Fatal error:  Uncaught Com\Tecnick\Pdf\Font\Exception: unable to read file: helvetica.json in /var/www/html/test-uw/tc-lib-pdf/vendor/tecnickcom/tc-lib-pdf-font/src/Stack.php:183
Stack trace:
#0 /var/www/html/test-uw/tc-lib-pdf/vendor/tecnickcom/tc-lib-pdf/examples/index.php(46): Com\Tecnick\Pdf\Font\Stack->insert()
#1 {main}
  thrown in /var/www/html/test-uw/tc-lib-pdf/vendor/tecnickcom/tc-lib-pdf-font/src/Stack.php on line 183
[Fri Jan 26 09:55:16 2024] 10.0.2.121:49293 [500]: GET / - Uncaught Com\Tecnick\Pdf\Font\Exception: unable to read file: helvetica.json in /var/www/html/test-uw/tc-lib-pdf/vendor/tecnickcom/tc-lib-pdf-font/src/Stack.php:183
Stack trace:
#0 /var/www/html/test-uw/tc-lib-pdf/vendor/tecnickcom/tc-lib-pdf/examples/index.php(46): Com\Tecnick\Pdf\Font\Stack->insert()
#1 {main}
  thrown in /var/www/html/test-uw/tc-lib-pdf/vendor/tecnickcom/tc-lib-pdf-font/src/Stack.php on line 183

nicolaasuni:

nicolaasuni:
Barcodes are already supported in the new tc-lib-pdf library.
If you need features of this library that have not been ported yet (HTML/CSS/SVG/Javascript), please use the tc-lib-barcode directly to generate the barcode and render it via SVG for example.

Issue #702, opened 2024-03-26
Update comment for renamed serializeTCPDFtagParameters


hakito:
Comment should be updated

https://github.com/tecnickcom/TCPDF/blob/d4adef47ca21c90e6483d59dcb9e5b1023696937/examples/example_049.php#L87

Btw. a breaking change should increase the version number to 7 when using semantic versioning. I had the version set to `^6.6` . This would [expect only non-breaking upgrades](https://getcomposer.org/doc/articles/versions.md#caret-version-range-).

nicolaasuni:
Thank you for reporting this. TCPDF 6 is now a legacy project and is officially **deprecated**, so no further fixes or features will be merged into this repository. As part of this transition, we are closing older issues. Please migrate to **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** and, if the same problem still occurs there, open a new issue with updated reproduction steps. For more context, see the pinned announcement: [Thank You - and the Future of TCPDF](https://github.com/tecnickcom/TCPDF/issues/867).

PR #703, opened 2024-03-26
Update comment in example_049.php

hakito:
Renamed method `serializeTCPDFtag`

fixexs #702


nicolaasuni:
This will be fixed in the next release as the comments are also not correct

williamdes:
https://github.com/tecnickcom/TCPDF/commit/820383ac064196c15c7a519f91e34914ec60702a

PR #705, opened 2024-03-29
Add-font-files

sajadshouli:


sajadshouli:
The Yekan font is the popular and beautiful font on the Persian web

I used the font in my project and It works fine. see the example

[sandwish_orders_pdf_1711734629jumksm9d3y-186_customer.pdf](https://github.com/tecnickcom/TCPDF/files/14807320/sandwish_orders_pdf_1711734629jumksm9d3y-186_customer.pdf)

nicolaasuni:
At this point I am not adding any more fonts in the default distribution as tc-lib-pdf is already able to print text and has more extensive fonts. You can always keep these locally.

sajadshouli:
I used the package in a lot of projects and for a font, I have to fork your project and the elibyy/tcpdf-laravel repo

PR #680, opened 2024-01-24
add setTcpdfLink method to allow enabling or disabling TCPDF meta link

it-can:
add setTcpdfLink method to allow enabling or disabling TCPDF meta link

PR #713, opened 2024-04-20
Add setTcpdfLink method to disable tcpdf footer

it-can:
Rework of #680


nicolaasuni:
This lib will be deprecated in favour of tc-lib-pdf.
I am trying to accept only essential PRs to keep this lib alive until all remaining features (HTML/SVG/CSS/Javascript) are ported.

Issue #665, opened 2023-11-13
Arial with $this->isunicode set to true


BSchuhmacher:
Hi
When using TCPDF with an old project (moved from fpdf to tcpdf) there is a problem using Arial as a font.  
In line 4277 there is a check for Arial but the font- familiy name is only set to helvetica if isunicode is false. I do not understand why Arial should only work with "isunicode=false".  
Uncommenting this check or setting the PDF creation to non unicode did not change anything as far as I could see.  

Is this an unnecessary (wrong) check or are there reasons not to change Arial to Helvetica if it is a non-Unicode document?

Regards
Bernd

rubo77:
I had the same problem:

Why this line?

https://github.com/tecnickcom/TCPDF/blob/951eabf0338ec2522bd0d5d9c79b08a3a3d36b36/tcpdf.php#L4285

This will replace arial with helvetica, which looks alike but not completely, see https://pimpmytype.com/arial-vs-helvetica/

nicolaasuni:
Thank you for reporting this. TCPDF 6 is now a legacy project and is officially **deprecated**, so no further fixes or features will be merged into this repository. As part of this transition, we are closing older issues. Please migrate to **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** and, if the same problem still occurs there, open a new issue with updated reproduction steps. For more context, see the pinned announcement: [Thank You - and the Future of TCPDF](https://github.com/tecnickcom/TCPDF/issues/867).

PR #733, opened 2024-07-19
allow arial as font although unicode is disabled

rubo77:
solves #665

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -4281,9 +4281,6 @@ class TCPDF {
 		}
 		// normalize family name
 		$family = strtolower($family);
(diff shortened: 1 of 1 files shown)


nicolaasuni:
Arial is a Unicode font, so when selected on a non-unicode doc an helvetica font is used instead.
That should not make any difference.
You can always experiment by changing font names.
Bear in mind the new tc-lib-pdf library has already a better font system and this lib will be deprecated.

rubo77:
>  That should not make any difference.

it makes a difference, look at https://pimpmytype.com/arial-vs-helvetica/

PR #741, opened 2024-08-28
Encrypt /DA string when document is encrypted

JedrzejczykRobert:
When document is encrypted then /DA string must be encrypted, without this Acrobat cannot allow fill form fields.


JedrzejczykRobert:
Steps to reproduce:

use code from examples\example_054.php

before "$pdf->Output('example_054.pdf', 'D');" insert 

$pdf->setProtection(array(), 'test', 'test', 3, null);

williamdes (approved):

nicolaasuni:
OK - Manually merged: https://github.com/tecnickcom/TCPDF/pull/755

PR #747, opened 2024-09-17
Improve quality of generated seed, avoid potential security pitfall

xelan:
* Try to use random_bytes() first if it's available
* Do not include the server parameters in the generated seed, as they might contain sensitive data

As all current usages of getRandomSeed() directly hash the seed, there should be no BC breaking changes.

The main source of entropy is more than enough on its own if random_bytes() or openssl_random_pseudo_bytes() are available.


williamdes (approved):

nicolaasuni:
OK - Manually merged: https://github.com/tecnickcom/TCPDF/pull/755

PR #765, opened 2024-11-26
improve SVG detection by checking for (mandatory) namespace and closi…

mcapinha-mxp:
This expands the regex used to detect SVGs by checking for the existence of the (mandatory) namespace attribute (`xmlns= http //www.w3.org/2000/svg`) and the closing `</svg>` tag.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -19055,7 +19055,7 @@ class TCPDF {
 				if ($imgsrc[0] === '@') {
 					// data stream
 					$imgsrc = '@'.base64_decode(substr($imgsrc, 1));
-					$type = preg_match('/<svg([^\>]*)>/si', $imgsrc) ? 'svg' : '';
+					$type = preg_match('/<svg\s+[^>]*\bxmlns\s*=\s*"http:\/\/www\.w3\.org\/2000\/svg"[^>]*>.*<\/svg>/is', $imgsrc) ? 'svg' : '';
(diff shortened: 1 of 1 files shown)


tcpdf.php
```diff
@@ -19055,7 +19055,7 @@ protected function openHTMLTagHandler($dom, $key, $cell) {
 				if ($imgsrc[0] === '@') {
 					// data stream
 					$imgsrc = '@'.base64_decode(substr($imgsrc, 1));
-					$type = preg_match('/<svg([^\>]*)>/si', $imgsrc) ? 'svg' : '';
+					$type = preg_match('/<svg\s+[^>]*\bxmlns\s*=\s*"http:\/\/www\.w3\.org\/2000\/svg"[^>]*>.*<\/svg>/is', $imgsrc) ? 'svg' : '';
```
williamdes, line 19058: can the xmlns be removed ? or are we 100% sure it will exist in all svg files ?

mcapinha-mxp: Hi @williamdes ,

From https://www.w3.org/TR/SVG2/struct.html#Namespace , my understanding is that when an image is inlined with its base64 representation, it should include the namespace as this is akin to the file being loaded externally.
The namespace can be omited for inline SVG code, but that's not the case here.

With that said, I agree that we can skip checking for the namespace.
I've updated the PR.

Thanks for taking the time to look into this.

williamdes (approved):

PR #775, opened 2025-01-01
suporte new font 'sansserif'

gamadoleo:


williamdes:
From what source files did you create the files ?

nicolaasuni:
Thank you for your contribution but adding proprietary fonts in not allowed.
Even if it was allowed this library is in maintenance-mode only as it should be replaced by tc-lib-pdf.

PR #784, opened 2025-01-30
feat(Fonts): Added new fonts on TCPDF

Carlos-hub:
Added fonts to the TCPDF converter file using http://www.fpdf.org/makefont/


williamdes:
This PR will never be merged in such a state (git history is a mess), and maybe not merged at all.
See: https://github.com/tecnickcom/TCPDF/pull/775#issuecomment-2614430070

nicolaasuni:
Sorry, I have to reject this as the PR is in a avery poor state with 414 file changes!
This library is in a kind of support-only mode.
A new replacement library is avaliabel even if it doesn't have yet html-kind features.
Please have a look instead at the tc-lib-pdf library and how to use fonts there.

Issue #684, opened 2024-02-02
How to output a PDF/A-3 file using the TCPDF library


f8solution:
Hello,

I am able to use the library correctly and generate a PDF file. However, if I understand correctly, the TCPDF library does not currently support direct output in the PDF/A-3 format (only PDF/A1-b).

Is an update planned to incorporate the choice of the output format?

If not, do you know how I can convert the file after the output command, using PHP to achieve PDF/A-3?

php code
// Save the PDF in the temporary folder
$err .= $pdf->Output($filePath, 'F');

Thank you in advance for your answer.

Best regards,

akrassler:
Hello,

that would be really interesting for me, too.
Are there any plans to implement PDF/A-3 compatibility?

Regards,
alex

jugantner2:
I am interested, too in PDF/A-3 in order to generate e-invoice compliant to Zugpferd.
Thanks to the team developing that cool lib.

williamdes:
https://github.com/tecnickcom/TCPDF/pull/789 is merged
@f8solution please close this one?

PR #789, opened 2025-02-12
Embedded files support (Factur-X 1.07 / ZUGFeRD 2.3)

### LINKED ISSUE #578: How to add a file into the PDF, without adding annotation?
Hello,

To create an invoice, using the factur-x format, I need to embed an XML file into the PDF. The only way, i've founded was to add an annotation:

```
$pdf->Annotation(0,0,0,0, 'invoice', array(
    'Subtype' => 'FileAttachment', 
    'Name'    => 'PushPin', 
    'FS'      => 'factur-x.xml'
));
```

But:
1) that adds an useless annotation
2) the file appears twice (in the file list) when I open the PDF with Accrobat Reader

Is there a way to simply add the file, without annotation?

Thx
Th.

ghost:
This PR should get merged after #788 - fixes #578, #684

The changes proposed in this PR allow the generation of compliant Factur-X 1.07 / ZUGFeRD 2.3 pdfs that embed additional invoice information as XML.

To achieve this one additional insertion point inside the XMP section for PDF/A extension has been added as well as methods to directly insert embedded files without the need to use annotations.

The last required change was to include all embedded files in the catalog under /AF to pass PDF/A-3b validation.

The pdf generated via example_068.php will pass the online validator for FacturX found here
https://services.fnfe-mpe.org/
Snip20250212_3
as well as https://www.portinvoice.com/ to validate it as Zugferd 2.3 compatible pdf
image


williamdes:
Very nice!

williamdes (approved):

williamdes:
For reference on Dompdf Factur-x support is near to work. Could you help me with the last VeraPDF error?
https://github.com/dompdf/dompdf/issues/3443

fooman:
Thanks @williamdes - updated the description. 

In regards to dompdf I am unfortunately not sure either what the error entails. One thing which tcpdf does when in pdfa mode is the forced embedding of fonts even for core fonts like https://github.com/tecnickcom/TCPDF/blob/main/fonts/pdfacourier.php and it also suppresses font subsetting.

nicolaasuni:
Please trey to add this instead on the tc-lib-pdf project.

williamdes:
> Please trey to add this instead on the tc-lib-pdf project.

Could you re-consider for this one?
this feature is very critical for all the tcpdf current users that want to get Factur-X compliance done

AdrienWeber:
> > > Please trey to add this instead on the tc-lib-pdf project.
> > 
> > 
> > Could you re-consider for this one? this feature is very critical for all the tcpdf current users that want to get Factur-X compliance done
> 
> Yes, could you please re-consider? I can confirm that I check this PR two/three times a week, because we need the ability to embed xml files complaint for ZUGFeRD...

Yes please, this would help a lot.

nicolaasuni (approved):

Issue #819, opened 2025-07-02
Compression can't be enabled for PDF/A compliant documents


schlotzz:
**Description:**
I've tried to enable compression for documents with PDF/A standard, but was not able to do so. The method _setCompression()_ disables compression as soon _pdfa_mode_ is active.
```
public function setCompression($compress=true) {
	$this->compress = false;
	if (function_exists('gzcompress')) {
		if ($compress) {
			if ( !$this->pdfa_mode) {
				$this->compress = true;
			}
		}
	}
}
```

**Issue:**
Compression cannot be enabled for PDF/A. As stated on [Wikipedia](https://en.wikipedia.org/wiki/PDF/A), only LZW compression is forbidden within PDF/A standard due to former intellectual property constraints. DEFLATE (a.k.a. "/Filter /FlateDecode") as it is used within TCPDF, is permitted as an open and patent-free standard.

**Proposed Solution:**
Compression should be allowed for all PDF/A-modes.

I already tested some code changes succesfully and could offer a pull request if wanted.

williamdes:
Can it be closed now that #820 is merged ?

PR #820, opened 2025-07-02
Enabled compression for PDF/A-3

schlotzz:
Enables compression for PDF/A according to Issue #819

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -2910,9 +2910,7 @@ class TCPDF {
 		$this->compress = false;
 		if (function_exists('gzcompress')) {
 			if ($compress) {
-				if ( !$this->pdfa_mode) {
-					$this->compress = true;
-				}
+                $this->compress = true;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


williamdes (commented):
@nicolaasuni you know the spec better
is this correct?

nicolaasuni:
The compression depends on the PDFA level.
If I remember correctly this is already addressed in the new tc-pdf-lib library.
Please have a look there...

tcpdf.php
```diff
@@ -2910,9 +2910,7 @@ public function setCompression($compress=true) {
 		$this->compress = false;
 		if (function_exists('gzcompress')) {
 			if ($compress) {
-				if ( !$this->pdfa_mode) {
-					$this->compress = true;
-				}
+                $this->compress = true;
```
nicolaasuni, line 2913: This shoudl depend on the PDF/A level

schlotzz:
Please have a look at my issue #819. As stated on [Wikipedia](https://en.wikipedia.org/wiki/PDF/A), only LZW compression is forbidden within PDF/A standard due to former intellectual property constraints. DEFLATE (a.k.a. "/Filter /FlateDecode") as it is used within TCPDF, is permitted as an open and patent-free standard.

As currently only DEFLATE is used, compression should be allowed for all PDF/A-modes.

nicolaasuni (approved):

schlotzz:
Thanks for your effort and merge 👍

PR #822, opened 2025-07-18
first push

nfceek:


nicolaasuni:
We sincerely appreciate the continued interest in this project, even after more than 25 years. A modern, fully functional version now exists: [tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf). 

Forking the legacy project may inadvertently divide efforts that could otherwise enhance the new tc-lib-pdf library. We kindly encourage you to consider contributing to the new project, which benefits from years of improvements and additional features.

Thank you for your support and understanding.

PR #833, opened 2025-10-06
Fix bad `text-align` from HTML source

j0k3r:
In case the `text-align` attribute is badly defined in the HTML, PHP raise an error: `Warning: Uninitialized string offset 0`

image

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -16918,7 +16918,7 @@ class TCPDF {
 							$dom[$key]['height'] = $dom[$key]['style']['height'];
 						}
 						// check for text alignment
-						if (isset($dom[$key]['style']['text-align'])) {
+						if (isset($dom[$key]['style']['text-align'][0])) {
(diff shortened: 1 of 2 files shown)


tcpdf.php
```diff
@@ -16918,7 +16918,7 @@ protected function getHtmlDomArray($html) {
 							$dom[$key]['height'] = $dom[$key]['style']['height'];
 						}
 						// check for text alignment
-						if (isset($dom[$key]['style']['text-align'])) {
+						if (isset($dom[$key]['style']['text-align'][0])) {
```
williamdes, line 16921:
```suggestion
						if (isset($dom[$key]['style']['text-align']) && isset($dom[$key]['style']['text-align'][0])) {
```
Will this avoid a crash if the text align key is missing ?

j0k3r: Yep: https://3v4l.org/gVpPk

williamdes: Awesome, thanks !
I checked for EOL versions too

nicolaasuni (approved):

PR #835, opened 2025-11-10
Fix PHP 8.5 deprecation for xml_parser_free

IndyIndyIndy:
Add php version check to xml_parser_free() for PHP 8.5 compatibility.

From PHP 8.0 onwards, this function does nothing. PHP 8.5 adds a deprecation warning for this function usage, which should be prevented.


williamdes (approved):
Looks okay

nicolaasuni (approved):

PR #838, opened 2025-11-25
Fix PHP 8.5 deprecation on curl_close()

mpesari:
`curl_close()` is noop since 8.0.0 and emits `E_DEPRECATED` since 8.5.0.

This changes the code to call `curl_close()` only on PHP < 8.0.0.

* https://www.php.net/curl_close
* https://github.com/php/php-src/blob/php-8.0.0/UPGRADING#L1016-L1020
* https://github.com/php/php-src/blob/php-8.5.0/UPGRADING#L416-L418

### DIFF
--- a/include/tcpdf_static.php
+++ b/include/tcpdf_static.php
@@ -1863,7 +1863,9 @@ class TCPDF_STATIC {
         curl_setopt_array($crs, $curlopts);
 		curl_exec($crs);
 		$code = curl_getinfo($crs, CURLINFO_HTTP_CODE);
-		curl_close($crs);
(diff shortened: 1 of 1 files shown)


MacGritsch:
Same should be done with imagedestroy. See attached patch.
[patch.diff.zip](https://github.com/user-attachments/files/23899128/patch.diff.zip)


There is code that checks for 'safe_mode', which was removed in PHP 5.4 (not fixed in patch).
Encryption/decryption may also not work but I have no code that uses this, so I have not tested/fixed it.

williamdes:
Fixes #850

PR #846, opened 2026-01-09
Remove blob image source from HTML

Lenny4:


williamdes:
Why did you close this?
Can you provide an example?

PR #849, opened 2026-02-02
Fixed: Font subsetting - incorrect placement of head checksum

jindrapetrik:
Currently, during font subsetting, head checksum is calculated and placed on offset 8 after head table start,
but this is incorrect and causes overwriting head.offset field!

The checksum should be placed in offset 4 after the tag 'head'.

This PR fixes this.

For me, this fixed PDF display error on Adobe Reader and or not diplaying proper characters in Edge PDF viewer.

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -1384,7 +1384,7 @@ class TCPDF_FONTS {
 		}
 		// set checkSumAdjustment on head table
 		$checkSumAdjustment = 0xB1B0AFBA - self::_getTTFtableChecksum($font, strlen($font));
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Can you provide a before and after pdf please?

jindrapetrik:
Yes, here:
[before.pdf](https://github.com/user-attachments/files/25009517/before.pdf)
[after.pdf](https://github.com/user-attachments/files/25009519/after.pdf)

Using code like this:
```php
      $pdf = new \TCPDF();
      $pdf->AddFont('Unifont', '', PDF_FONTPATH . 'unifont.php', true);
      $pdf->setFont('Unifont', '', 16);
      $pdf->AddPage();
      $pdf->Write(16, "ABC");
      $pdf->Output('out.pdf', 'D');
```
(I am using TCPDF 6.4.4, but the problem is probably the same in current.)

williamdes:
Thanks
@nicolaasuni ready for your review

PR #853, opened 2026-02-13
Fix null array offset in _destroy() for PHP 8.5

### LINKED ISSUE #852: PHP 8.5 deprecation: null array offset in _destroy() when file_id is null
### Description

In PHP 8.5, using `null` as an array offset is deprecated. The `_destroy()` method in `tcpdf.php` accesses `self::$cleaned_ids[$this->file_id]` without checking if `$this->file_id` is null first, triggering:

```
Deprecated: Using null as an array offset is deprecated, use an empty string instead
```

This happens at line 7891 (`self::$cleaned_ids[$this->file_id]`).

### Steps to reproduce

1. Use PHP 8.5
2. Generate any PDF — the deprecation warning is triggered in the destructor chain

### Suggested fix

Add an early return in `_destroy()` when `$this->file_id` is null, before accessing `self::$cleaned_ids`.

kolotov:
In PHP 8.5, using null as an array offset is deprecated.
The `_destroy()` method accesses `self::$cleaned_ids[$this->file_id]`
without checking if `$this->file_id` is null, triggering a deprecation warning.

Add an early return when `$this->file_id` is null.

Fixes: #852


williamdes (changes requested):
I am not sure this is correct, there is more code that should maybe be executed

kolotov:
You're right. I'll think about it some more.

renanrodrigo:
As this is just about unsetting `destroyall` when the file has already been cleaned, it's enough to check if there is any `file_id` at all. If it's null, the second part of the method can still unset the other properties if `destroyall` is set.

```
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -7882,7 +7882,7 @@
 	 * @since 4.5.016 (2009-02-24)
 	 */
 	public function _destroy($destroyall=false, $preserve_objcopy=false) {
-		if (isset(self::$cleaned_ids[$this->file_id])) {
+		if (isset($this->file_id) && isset(self::$cleaned_ids[$this->file_id])) {
 			$destroyall = false;
 		}
 		if ($destroyall AND !$preserve_objcopy && isset($this->file_id)) {
```

williamdes (approved):

kolotov:
@renanrodrigo  Good point, I have applied your suggested change.

nicolaasuni:
fixed on another PR

Issue #841, opened 2025-12-09
TCPDF now requires php_curl


MartinHusemann:
Commit aab43ab0a824e956276141a28a24c7c0be20f554  introduced a default curl options array CURLOPT_DEFAULT, but the code is not parsable without php_curl being available.

Without it, parsing fails like this:

    PHP Fatal error:  Uncaught Error: Undefined constant "CURLOPT_CONNECTTIMEOUT" in .../include/tcpdf_static.php:115
    Stack trace:
    #0 .../include/tcpdf_static.php(2511): [constant expression]()

williamdes:
Already declared as a composer requirement

https://github.com/tecnickcom/TCPDF/blob/7a2701251e5d52fc3d508fd71704683eb54f5939/composer.json#L26

zizzo81:
I don't use composer.

This could be simplified for people that use more rigorous environments like me just converting 
https://github.com/tecnickcom/TCPDF/blob/7a2701251e5d52fc3d508fd71704683eb54f5939/include/tcpdf_static.php#L114
and
https://github.com/tecnickcom/TCPDF/blob/7a2701251e5d52fc3d508fd71704683eb54f5939/include/tcpdf_static.php#L129
to functions that return the array, solves the problem.

Could you please consider a fix like this?

williamdes:
Please open a pull request, if you do minimal changes it can be accepted

zizzo81:
Done, created PR #854 
Thank you for the great work!

nicolaasuni:
Thank you for reporting this. TCPDF 6 is now a legacy project and is officially **deprecated**, so no further fixes or features will be merged into this repository. As part of this transition, we are closing older issues. Please migrate to **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** and, if the same problem still occurs there, open a new issue with updated reproduction steps. For more context, see the pinned announcement: [Thank You - and the Future of TCPDF](https://github.com/tecnickcom/TCPDF/issues/867).

PR #854, opened 2026-02-14
Decreased Curl dependency

zizzo81:
This modification makes possible to use TCPDF without forcing user to have the CURL extension enabled, provided you accept the limitation of not using any CURL-dependent functions.

### DIFF
--- a/include/tcpdf_static.php
+++ b/include/tcpdf_static.php
@@ -107,29 +107,33 @@ class TCPDF_STATIC {
 	public static $pageboxes = array('MediaBox', 'CropBox', 'BleedBox', 'TrimBox', 'ArtBox');
 
 	/**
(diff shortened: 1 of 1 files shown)


williamdes (approved):
Looks okay, thanks

nicolaasuni:
Thank you very much, @zizzo81, for the careful work to make TCPDF usable in environments without the cURL extension - converting the class constants that contain cURL constants into lazy accessor functions is a clean approach, and the companion issue #841 makes the problem very clear.

In **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)**, cURL is an optional dependency and the library degrades gracefully when it is not available. If you encounter a situation where that is not the case, feel free to open a PR or issue there.

PR #857, opened 2026-02-25
Ignore warnings about float to int casting on tcpdf_fonts

### LINKED ISSUE #856: Test fails on armhf with PHP 8.5 due to warning about casting values to `int`
When running the test suite on armhf and using PHP 8.5, a warning is displayed when some casts to int lead to overflows.
This was caught during the PHP 8.5 migration on Ubuntu. Here are the [logs of a test run](https://objectstorage.prodstack5.canonical.com/swift/v1/AUTH_0f9aae918d5b4744bf7b827671c86842/autopkgtest-resolute/resolute/armhf/t/tcpdf/20260223_123111_33caf@/log.gz) where you can read:
```
Warning: The float 2410464172741 is not representable as an int, cast occurred in /usr/share/php/tcpdf/include/tcpdf_fonts.php on line 941
```
These warnings are polluting the pdf output and causing the tests to fail. On Ubuntu, I've added a patch to just ignore the warnings and move on... I can open a PR with this patch, but I'd like to know if you have a better idea around the issue.

renanrodrigo:
The warning is being thrown in PHP 8.5, and it's breaking PDF generation in some cases, caught in tests executed on armhf systems.

Fixes: #856

----------------------------------------------

If you think masking this error may be dangerous or mask real situations which should be taken care of, feel free to reject/modify the approach.

I think this is safe because it's a warning which happens only in 32bit systems, where the cast is already present anyway and the overflows are happening already :)

### DIFF
--- a/include/tcpdf_fonts.php
+++ b/include/tcpdf_fonts.php
@@ -930,6 +930,12 @@ class TCPDF_FONTS {
 	 * @public static
 	 */
 	public static function _getTTFtableChecksum($table, $length) {
+		// Suppress warning about casting on armhf
+		if (PHP_INT_SIZE == 4) {
+			set_error_handler(function($errno, $errstr) {
+				return strpos($errstr, 'is not representable as an int') !== false;
+			}, E_WARNING);
+		}
 		$sum = 0;
 		$tlen = ($length / 4);
 		$offset = 0;
@@ -939,6 +945,9 @@ class TCPDF_FONTS {
 			$offset += 4;
 		}
 		$sum = unpack('Ni', pack('N', $sum));
+		if (PHP_INT_SIZE == 4) {
+			restore_error_handler();
+		}
 		return $sum['i'];
 	}
 
@@ -1382,9 +1391,18 @@ class TCPDF_FONTS {
 		foreach ($table as $data) {
 			$font .= $data['data'];
 		}
+		// Suppress warning about casting on armhf
+		if (PHP_INT_SIZE == 4) {
+			set_error_handler(function($errno, $errstr) {
+				return strpos($errstr, 'is not representable as an int') !== false;
+			}, E_WARNING);
+		}
 		// set checkSumAdjustment on head table
 		$checkSumAdjustment = 0xB1B0AFBA - self::_getTTFtableChecksum($font, strlen($font));
 		$font = substr($font, 0, $table['head']['offset'] + $offset + 8).pack('N', $checkSumAdjustment).substr($font, $table['head']['offset'] + $offset + 12);
+		if (PHP_INT_SIZE == 4) {
+			restore_error_handler();
+		}
 		return $font;
 	}
 


include/tcpdf_fonts.php
```diff
@@ -45,6 +45,12 @@
  * @version 1.1.1
  * @author Nicola Asuni - info@tecnick.com
  */
+
+// Suppress warning about casting on armhf
+set_error_handler(function($errno, $errstr) {
```
williamdes: Okay, I was not expecting such a patch
it does not fix it but hides the warning
I guess some debuting is needed to fix it for real?
Or this error suppression could be done around the problematic lines?

include/tcpdf_fonts.php
```diff
@@ -45,6 +45,12 @@
...
+    return strpos($errstr, 'is not representable as an int') !== false;
+}, E_WARNING);
```
williamdes: Aside from that, I would prefer not to keep a diff in Ubuntu
can you send your patch in Debian salsa?
so you can sync the version afterwards

renanrodrigo: Sure thing: once we agree on a good patch I can send it in salsa too (:

include/tcpdf_fonts.php
renanrodrigo: well yep.
This is ignoring the error just for the sake of the Ubuntu transition to PHP8.5, so we are not blocked, but the bug is still there - and it was there before, PHP is just complaining about it now...

I fear I don't have enough context to have a good solution for this issue soon, but I can look further in the future.

One thing I ended up doing for other packages is limiting it to 32bit PHP (by wrapping it in a check on `PHP_INT_SIZE`)

williamdes:
> (by wrapping it in a check on PHP_INT_SIZE

Good idea indeed
Plus moving your patch around the problematic lines would be better

renanrodrigo:
@williamdes 

So, from what I can see, there is a checksum calculation that is overflowing - I am not sure on how to fix that the proper way, for I don't know exactly where this checksum is being used later (not a fonts specialist :|). 

I updated this PR with the same code I applied in Ubuntu, and I see you have it in salsa too.
Let me know if there is something else I can help with here.

Thanks!

williamdes:
Thank you very much!
Yes that will be easier for someone else to debug this. I'd say there is barely a chance that it occurs but we will see

~~You can sync my Debian version when I upload it?~~
Edit: nevermind, I did read https://code.launchpad.net/~rr/ubuntu/+source/tcpdf/+git/tcpdf/+merge/500987

nicolaasuni:
Thank you very much, @renanrodrigo, for the PHP 8.5 armhf / 32-bit float-to-int casting fix in `tcpdf_fonts` - catching platform-specific overflow warnings on 32-bit ARM builds is exactly the kind of careful maintenance that helps downstream packagers, and it is genuinely appreciated.

The font subsystem in **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)** has been rewritten with proper integer typing throughout. If you encounter similar warnings there, feel free to open a PR or issue.

renanrodrigo:
Hey @nicolaasuni thanks for the heads up, and the path forward looks good btw (:

Issue #796, opened 2025-04-02
Deprecations message when calling writeHTMLCell() with specifics HTML codes


sdespont:
When I call this function with the following comments, deprecation are trigged `$tcPdf->writeHTMLCell(190, 70, null, null, $comment, 0, 0);`

`$comment = "<p><strong>sdfas </strong>df</p>";`
Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated
https://github.com/tecnickcom/TCPDF/blob/f67b761b61f2370a9000d98c3b9111284544f722/tcpdf.php#L18547

--------

`$comment = "<p>A <strong>B</strong>, C.<br /><br />D<strong> E,</strong> F</p>";`
Deprecated: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated
https://github.com/tecnickcom/TCPDF/blob/f67b761b61f2370a9000d98c3b9111284544f722/tcpdf.php#L17695

--------
`$comment = "<p>A <strong>B</strong>, C.<br /><br />D<strong> E,</strong> F</p>";`
Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated 
https://github.com/tecnickcom/TCPDF/blob/f67b761b61f2370a9000d98c3b9111284544f722/tcpdf.php#L6395

Could you have a look ?

littlepackage:
Hi! I cannot reproduce this with the $comment values you gave when I use the latest version of TCPDF, using PHP versions 7.4 and 8.4.

```
$comment = "<p><strong>sdfas </strong>df</p>";
$tcPdf->writeHTMLCell(190, 70, null, null, $comment, 0, 0);
```

works.

```
$comment = "<p>A <strong>B</strong>, C.<br /><br />D<strong> E,</strong> F</p>";
$tcPdf->writeHTMLCell(190, 70, null, null, $comment, 0, 0);
```

works, too. If you are still having a problem, what version of TCPDF and PHP you are using, and can you provide more lines of your code and note if you have changed any config settings.

sdespont:
@littlepackage yes it "works" but there are deprecation notices to resolve.

```
<?php
error_reporting(E_ALL);
ini_set('display_errors', '1');
require './vendor/tecnickcom/tcpdf/tcpdf.php';

$tcPdf1 = new TCPDF('P', 'mm', 'A4', true, 'ISO-8859-1');
$tcPdf1->AddPage();
[...]

PR #863, opened 2026-04-08
Resolved deprecation notice regarding deprecated strlen() of null-value in Write() function

samy-rd:
As mentioned in #796 sometimes the following deprecation notice occurs while using `strlen()` on a null value (non-string) in the `tcpdf->Write()` function if its param `$txt` is null/not given. 

`PHP message: PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in .../tcpdf.php on line 6425;`

This small fix should handle cases where `$txt` is null/not set.


williamdes (approved):

nicolaasuni:
Thank you very much, @samy-rd, for the fix for the `strlen(null)` deprecation notice in `Write()` - keeping the library clean on PHP 8.1+ with strict deprecation reporting is important for many projects, and it is genuinely appreciated.

Null-safety throughout the codebase is enforced by PHPStan in **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)**, so deprecation notices of this kind do not arise. If you encounter a related issue there, feel free to open a PR or issue.

Issue #767, opened 2024-11-29
Rendering of SVG -> PDF issue


BilboTav:
I use tcpdf to render SVG template into PDF for my CV and certain snippet is badly rendered on newest version:

6.7.5 (looks fine)
![image](https://github.com/user-attachments/assets/0c6bc0ee-813f-4398-b6b0-2ec01d828996)

vs

6.7.7 (edge point is ugly rendered)
![image](https://github.com/user-attachments/assets/0698b3cb-ca42-4810-8d80-7c1ad398cd92)

SVG part is this:

```
<path fill="{$color}"
              d="m 49.854628,14.022875 c 14.402622,0 26.928265,8.131049 33.359621,20.109374 l 6.74e-4,-3.41e-4 c
              0,0 0.717926,1.310865 1.503538,1.396184 0.785608,0.08532 1.081936,-1.855893 1.081936,-1.855893 l 0.002,6.83e-4 C
              87.752132,14.926248 104.28866,14.852189 104.28866,14.852189 h 14.50128 91.2102 V 0.00504815 H 0.00158345 V 84.119175
              H 6.9319861 c 5.2194399,0 4.9918029,-6.812334 4.9918029,-6.812334 v -23.359404 0 c -0.01751,-0.482916 -0.02694,-0.967535
              -0.02694,-1.454546 0,-21.246185 16.994158,-38.470016 37.957778,-38.470016 z" />
```

I am just giving heads up - maybe this fluke of mine is a sign of some other issues with regards of rendering SVG with Your library ?

Thanks.

williamdes:
Can you try to revert some of the changes in https://github.com/tecnickcom/TCPDF/compare/6.7.5...6.7.7 to find the one that is wrong ?

BilboTav:
> Can you try to revert some of the changes in [6.7.5...6.7.7](https://github.com/tecnickcom/TCPDF/compare/6.7.5...6.7.7) to find the one that is wrong ?

Hello, sorry for late response - issue starts with 6.7.7, 6.7.6 is fine. Thanks You!

grizzm0:
Similar issue here. Reverting to 6.7.6 did the trick. Will do that for now.

SVG vs rendered PDF
![logo](https://github.com/user-attachments/assets/80749d38-30dc-4298-af94-30f6e94212dc)

![Screenshot 2025-01-13 at 08 18 14](https://github.com/user-attachments/assets/232ba215-004b-454c-913e-ddf7cbd02524)

[...]

PR #864, opened 2026-04-14
Fix SVG curve parameter extraction once more

ferdiusa:
The switch from `preg_match_all` to `preg_split` in PR #855 broke parsing of "optimized" svg paths. In these, the "-" can be used as delimiter between numbers as well.

See Issue #767 for report by @tmhf.

The  regular expression suggested in this PR reverts back to token based behaviour while keeping the fix for "e" exponents from the original issue.

### DIFF
--- a/tcpdf.php
+++ b/tcpdf.php
@@ -23726,10 +23726,8 @@ class TCPDF {
 			$params = array();
 			if (isset($val[2])) {
 				// get curve parameters, see https://github.com/tecnickcom/TCPDF/issues/767
-				$rawparams = preg_split('/([\,\s]+)/si', trim($val[2]));
-				$rawparams = array_filter($rawparams, function($p) {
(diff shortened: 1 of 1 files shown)


williamdes (approved):
thank you

nicolaasuni:
Thank you very much, @ferdiusa, for the SVG curve parameter fix - the negative-number-as-delimiter edge case in `preg_split` on SVG path data is a subtle and tricky bug, and it is genuinely appreciated that you tracked it down once more.

SVG rendering has been redesigned in **[tc-lib-pdf](https://github.com/tecnickcom/tc-lib-pdf)**. If you are able to reproduce the path parsing issue there, feel free to open a PR or issue. If you need any help adapting your fix to the new library, please open a discussion and I will do my best to assist.

